PR #9, opened 2020-03-12
Do not allow individual capacity counts to be higher than total + ref…

Gerrit91:
…actor.

The counter-intuitive sum was caused by a machine that had issues and that was allocated at the same time, which was leading to the sum of the individual counts being higher than total.
The count can still be lower than the total count because there can be machine that are for example "preparing" at the moment. They are neither allocated nor free nor faulty.

This is possible right now:

```
PARTITION       SIZE            TOTAL   FREE    ALLOCATED       FAULTY 
fel-wps101      c1-xlarge-x86   92      37      50              6       
fel-wps101      s2-xlarge-x86   12      12      0               0     
```

92 != 37+50+6

### DIFF
--- a/cmd/metal-api/internal/service/partition-service.go
+++ b/cmd/metal-api/internal/service/partition-service.go
@@ -305,7 +305,7 @@ func (r partitionResource) partitionCapacity(request *restful.Request, response 
 	}
 }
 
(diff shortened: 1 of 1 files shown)


majst01:
A test would be nice

Gerrit91:
The previous "test" is still there and succeeds because the behaviour did not change. You mean a specific test, which would go wrong with the previous implementation?

Issue #21, opened 2020-03-13
Mitigate "machine was allocated without proper switch connections"


majst01:
There are two possibilities to get into a state where you have a machine that you cannot reach over the network after the allocation:

1. We register a new switch at the metal-api (metal-core starts for the first time and registers), but machines are already in waiting state (which can happen after a wrong update sequence or broken switch)
1. You start a machine which has a blade switch in between (like t1-small) where LLDP cannot discover the connections to the leaf switches

In both cases, we cannot find out to which switch a machine is connected to.

This can lead to the following failure state:

- When you allocate a machine that is not in the switches machine connections 
- The machine starts to boot
- The machine will not be enslaved into a VRF
- The machine will not be reachable from external networks

Can we prevent this state? As this is actually confusing... the resulting machines are unusable for a user.

For scenario (1) you can get the switch connection after rebooting the machine and everything would be fine.

---
Both problems can be mitigated an assertion like this: the machine report should fail if there are not two switches visible from the machines.
This will cause the report to fail more often and the t1-small servers won't get to the waiting state any more.

---

To be honest, it is not so likely to get into this state. The last time this happened was because we updated the metal-core, the metal-api and wiped the rethinkdb. However, it's better for the robustness if we prevent these states anyway as they are possibly easy to prevent.

The problem is: The metal-api does not care if there are two switch connections to the machine or not. It will allow machine allocation without this condition fulfilled. The metal-hammer could actually report some wild stuff about switch neighbors to the metal-api, the api would say "fine" and when you allocate it, you would end up with an unusable machine. And this is what happened: The "machine connections" got lost because we had new switches registered at the api, but the machines behind the switches were already in the waiting state. The metal-api should at least validate if it is actually able to construct a proper switch configuration before allowing machine allocation.

--

Ideally, such a machine should not even be able to enter the wait table. This would cause a reboot of the machine re-reporting the connections + not having a user allocate such a machine.

majst01:
@Gerrit91 was #31 related to this ? cant remember why, maybe @mwindower has some helpful input as well

Gerrit91:
IMHO we should add a validation of the reported registration data and prevent the metal-hammer to enter the wait phase when for example the neighbor condition cannot be verified from the metal-api perspective.

Gerrit91:
It was not related to #31.

mwindower:
It is related to #31 because connectMachineWithSwitches of the switch service is called during machine registration.
[...]

PR #31, opened 2020-03-13
Support switch replacement

### LINKED ISSUE #30: Support switch replacement
Therefore we need:

 - an API endpoint to edit switches (introduce a `mode` attribute for it that may be `operational` or `replace`)
 - changes for the `/register` endpoint to  
   - map mac addresses for the switch ports
   - adopt the vrf settings and machine connections that are present at the twin brother switch
 
Operational steps to replace a switch:
- replace the switch physically
- put the switch into `replace` mode at the metal-api - probably with `metalctl switch edit <id>` and adding `mode: replace`
- run the playbook for leaf switches: bake and deploy but make sure that interfaces get not immediately connected to the PXE-bridge by setting the interfaces dictionary of the respective leaf switch to `[]`
- the switch will initiate a switch registration (`/register`)
- metal-api will adopt the configuration from the old, broken switch to the new switch
- if the replacement was successful, metal-api will put the switch into `operational` mode again

mwindower:
Closes #30


cmd/metal-api/internal/service/switch-service.go
```diff
@@ -166,16 +207,17 @@ func (r switchResource) registerSwitch(request *restful.Request, response *restf
 			return
 		}
 
-		// TODO: Broken switch: A machine was registered before this new switch is getting registered
-		// It needs to take over the existing connections from the broken switch or something?
-		// metal/metal#28
-
 		returnCode = http.StatusCreated
+	} else if s.Mode == metal.SwitchReplace {
```
majst01, line 211: Ha, i like this

cmd/metal-api/internal/service/switch-service.go
```diff
@@ -166,16 +207,17 @@ func (r switchResource) registerSwitch(request *restful.Request, response *restf
 			return
 		}
 
-		// TODO: Broken switch: A machine was registered before this new switch is getting registered
```
Gerrit91, line 169: I think the comment does not fully relate to the broken switch story due to the way you are solving the issue here. However, it seems the comment still contains a valid point. What happens when there is only one switch registered, a machine registers and after that the second switch comes up? In my opinion we should validate the two-neighbor condition when a machine registers and also when a machine gets allocated, which relates to https://github.com/metal-stack/metal-api/issues/21.

Maybe just remove "Broken switch" from the comment?

mwindower: Yes, this is an issue. We now assume stronger constraints during the switch replacement then for the machine registration!

I would suggest to add these constraints to the machine registration:
* there must be two distinct switches that have connections to the machine
* these two switches must have "equal" machineConnections with respect to the switch port names

This would also help with #21

mwindower: I added operational steps needed to replace the switch in https://github.com/metal-stack/metal-api/issues/30

Gerrit91: Sounds good.
Regarding operational steps... how do I put the switch into replace mode as an operator? Will it become a metalctl command?

mwindower: Just added the additional constraints for the machine registration as well.

`metalctl switch apply / edit <ID>` is probably sufficient to put a switch into `replace` mode or back to normal `operational` mode.

A flag like `--mode` is not that appropriate as we just have those kind of flags for creation of entities but, as of now, not for editing.

cmd/metal-api/internal/service/switch-service.go
```diff
@@ -77,6 +77,16 @@ func (r switchResource) webService() *restful.WebService {
...
 		DefaultReturns("Error", httperrors.HTTPErrorResponse{}))
 
+	ws.Route(ws.POST("/").
+		To(admin(r.updateSwitch)).
+		Operation("updateSwitch").
+		Doc("updates a switch. if the switch was changed since this one was read, a conflict is returned").
+		Metadata(restfulspec.KeyOpenAPITags, tags).
+		Reads(v1.SwitchUpdateRequest{}).
+		Returns(http.StatusOK, "OK", v1.SwitchResponse{}).
+		Returns(http.StatusConflict, "Conflict", httperrors.HTTPErrorResponse{}).
```
Gerrit91, line 87: The `updateSwitch` function does not return a conflict status code explicitly, I think this line can be removed. Should be covered by default response

mwindower:
won't merge yet because integration test fails

majst01:
needs merge from master, i think we should not forget about this

mwindower:
I will merge this since the integration tests are really broken and this is not caused by this PR.

majst01:
@mwindower do we need a metalctl adoption here as well ?

PR #33, opened 2020-03-16
Json tags

kolsa:


Gerrit91:
I don't get this to be honest. :(

kolsa:
`metal.Machine` for example is part of `metal.MachineEvent` that will be published by NSQ and therewith marshalled to JSON. Although explicit JSON tags are not necessary at all due to implicit name resolution, it is required in the context of the current reinstall implementation to distinguish between the two fields `metal.MachineAllocation#Reinstall bool` and `v1.machineAllocation#Reinstall *v1.MachineReinstall`. Apart from that, to be explicit is always better than to be implicit.

majst01:
I think the main concern of @Gerrit91 is that we tried to distinguish between internal and external API´s and with this we solve a problem with was caused by a mixture of the above design. Send internal structs to NSQ is the root cause of the issue. 

Should we solve this issue the other way round ? WDYT ?

Gerrit91:
Okay, I think I got it. So, we never decoupled NSQ messages from the data store, right?

That's a little unfortunate and I share my concerns with @majst01. The database representation should be loosely coupled from the rest of the code for several reasons, i.e. API changes decoupled from database migrations, possible postgres data store implementation in the future, ...
In my opinion, NSQ messages should also use the API v1 structs. It requires a little refactoring though. Should we make another issue for that?

majst01:
If we remove the NSQ messages for switch events, only a few are left. Yes lets create another issue describing this. But we are fine with this change, arent we ?

Gerrit91:
We can have the tags there for now, but the issue that removes the NSQ messages from there needs to remove the tags again because they are not of any use and therefore confusing.

Thanks for clarification. :)

PR #43, opened 2020-03-23
Fix pr pipeline

Gerrit91:
At the moment, a PR builds an image with the tag `merge`...


Gerrit91:
I think this should be in a working state right now... it looks a little iffy, but Github actions does not easily support setting environment variables in between steps.

Gerrit91:
I can add this to metal-core and metal-hammer if I know it works (after the next release of this).

.github/workflows/tagged.yaml
```diff
@@ -14,15 +14,20 @@ jobs:
...
-    - name: Branch name
-      run: echo running on branch ${GITHUB_REF##*/}
     - name: Lint
       uses: actions-contrib/golangci-lint@master
       with:
         args: run
-    - name: Build the Docker images
+    - name: Export tag on pull request
+      run: | 
+        echo "::set-env name=TAG_NAME::${GITHUB_HEAD_REF##*/}"
```
Gerrit91: I stole that from https://github.com/actions/starter-workflows/issues/68#issuecomment-552074596

majst01: If is ugly, why not a separate workflow with a matching "on" statement ?

Gerrit91: Mh. Because I named it "tagged" image. But I can also make a workflow "release" and one "pull_request". Maybe that's clearer then.

PR #46, opened 2020-03-24
do not publish any model object at all, IDs are sufficient

kolsa:


kolsa:
The only receiver of machine commands is metal-core, and metal core is only interested on the machine IDs as well as the params. Thus we can narrow the whole struct to only these values and therewith omit the v1 vs. metal model conflict

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -1525,7 +1525,7 @@ func freeMachine(ds *datastore.RethinkStore, publisher bus.Publisher, ipamer ipa
 		return err
 	}
 
-	deleteEvent := metal.MachineEvent{Type: metal.DELETE, Old: m}
+	deleteEvent := metal.MachineEvent{Type: metal.DELETE, OldMachineID: m.ID}
```
majst01: is this the only occurrence of MachineEvent, if yes, great

kolsa: Yes :)

PR #47, opened 2020-03-24
implement async free machine

ulrichSchreiner:
implement async deletion of machines and connected ressources


majst01:
So this depends on https://github.com/metal-stack/metal-lib/pull/7 where bus.Endpoints is implemented.
Looks nice and clean from my POV.

My main concern with this approach is the concept of releasing the machine/ip  of the same context.  If something bad happens, and this will happen as we see actually, a machine/ip could be deleted with this approach if it was already allocated by another action.

But i dunno have a concrete idea howto cope with this.

ulrichSchreiner:
it would be great if an `IP` struct would have a synthetic UUID-field. in this case we could delete a specific entity of an IP-address and not the specific address (which could already be allocated by another machine).

a second uuid-field would be ok, the `ID` could still be the address itself. but on every allocation this unique field could be filled with a new uuid. in this case the deletion code can check if the entity of the persistent IP-adress has the same uuid value as the one to be deleted.


WDYT?

majst01:
I already was thinking about this already, is kind of related to a issue in go-ipam for support overlapping Prefixes/CIDRs which would be required to be able to give all kubernetes clusters the same nodeCIDR for example.
So this is a two fold question:
- ID + UUID in the metal-api
- ID + UUID in go-ipam
for the latter, i dunno if this helps ?

ulrichSchreiner:
i only thought to extend `metal-api` to make sure not to delete an IP multiple times if it is allocated between the deletions. 
if the same problem exists in `go-ipam` then the ipam should generate a unique ID for an allocated IP and `metal-api` should store this UUID too. in this case a deletion would hit the exact entity in both persistent stores.

but the problem will be much bigger if you want overlapping prefixes/cidrs. in `metal-api` the `IPAddress` is the primary key in the IP-table. this has to be reworked.

so perhaps we should think about using a synthetic `id` in the IP struct and set the `IPAddress` as a normal (non-unique) field.

ulrichSchreiner:
wait for https://github.com/metal-stack/metal-lib/pull/8 to be merged and a new version is released so i can update the `go.mod`

majst01:
@ulrichSchreiner can you please move this to a branch of metalstack/metal-api because otherwise we would not get a docker image from this MR.

ulrichSchreiner:
replaced by #66

PR #48, opened 2020-03-25
Not locked but reserved machines should not be counted a "free".

Gerrit91:


majst01:
But locked machines shouldn't be counted as free also ?

Gerrit91:
You are right. I thought locked machines had to be allocated. But it's not the case.

PR #51, opened 2020-04-07
WIP: Semantic images

majst01:


cmd/metal-api/internal/datastore/image.go
```diff
@@ -35,3 +47,146 @@ func (rs *RethinkStore) DeleteImage(i *metal.Image) error {
...
+	os := imageParts[0]
+	version := strings.Join(imageParts[1:], "")
+	v, err := semver.NewVersion(version)
+	if err != nil {
+		return "", nil, err
+	}
+	return os, v, nil
+}
+
+func sortImages(images []metal.Image) []metal.Image {
```
ulrichSchreiner, line 175: i don't get if this is a correct sorter for different images. you sort the slice twice, first by the OS, than again by the version.
when you have a list of images:
~~~
ubuntu-20.04.20200401
ubuntu-20.04.20200402
alpine-20.04.20200401
alpine-20.04.20200402
~~~
then after the first sort you have:
~~~
alpine-20.04.20200401
alpine-20.04.20200402
ubuntu-20.04.20200401
ubuntu-20.04.20200402
~~~
and after the seconds sort:
~~~
alpine-20.04.20200401
ubuntu-20.04.20200401
alpine-20.04.20200402
ubuntu-20.04.20200402
~~~
???

i might be wrong ... but what's the first sort for if your second resorts and orders by version?
EDIT: ok i see, as you use the `Stable` sort, you maintain the first sorting order for equal OS'es. couldn't this be done with one single compare-function which does the semver-compare and if both are equal, it returns the os-comparison?

cmd/metal-api/internal/metal/image.go
```diff
@@ -3,15 +3,61 @@ package metal
...
+	Version string `rethinkdb:"version" json:"version"`
+	// ExpirationDate defines the time in the future, when this image is not considered for machine allocations anymore
+	ExpirationDate time.Time `rethinkdb:"expirationDate" json:"expirationDate"`
+	// Classification defines the state of a version (preview, supported, deprecated)
+	// only informational, no action depending on the classification done
+	Classification VersionClassification `rethinkdb:"classification" json:"classification"`
 }
 
+// VersionClassification is the logical state of a version according to
+// https://github.wdf.sap.corp/kubernetes/kube-docs/wiki/Versioning-Policy
```
ulrichSchreiner, line 26: on my computer this link is dead

cmd/metal-api/internal/metal/image.go
```diff
@@ -3,15 +3,61 @@ package metal
...
+	ClassificationSupported VersionClassification = "supported"
+	// ClassificationDeprecated indicates that a patch version should not be used anymore, should be updated to a new version
+	// and will eventually expire.
+	// Every version that is neither in preview nor supported is deprecated.
+	// All patch versions of not supported minor versions are deprecated.
+	ClassificationDeprecated VersionClassification = "deprecated"
+)
+
+var versionClassifications = map[string]VersionClassification{
+	"preview":    ClassificationPreview,
```
ulrichSchreiner, line 45: if the strings are always equal you could do 
~~~
string(ClassificationPreview): ClassificationPreview
~~~
but this is ok too.

cmd/metal-api/internal/service/image-service.go
```diff
@@ -159,14 +160,38 @@ func (ir imageResource) createImage(request *restful.Request, response *restful.
 		features[ft] = true
 	}
 
+	os, v, err := ir.ds.GetOsAndSemver(requestPayload.ID)
+	if checkError(request, response, utils.CurrentFuncName(), err) {
+		return
+	}
+
+	expirationDate := time.Now().Add(time.Hour * 24 * 90)
```
ulrichSchreiner, line 168: shouldn't this duration be a constant?

cmd/metal-api/internal/datastore/image.go
majst01: Yes, there was even a BUG here, modified implementation and tests to cover this and have a more clear implementation.

cmd/metal-api/internal/datastore/image.go
```diff
@@ -35,3 +47,146 @@ func (rs *RethinkStore) DeleteImage(i *metal.Image) error {
 func (rs *RethinkStore) UpdateImage(oldImage *metal.Image, newImage *metal.Image) error {
 	return rs.updateEntity(rs.imageTable(), newImage, oldImage)
 }
+
+// DeleteOrphanImages deletes Images which are no longer allocated by a machine and older than allowed.
+// Always at least one image per OS is kept even if no longer valid and not allocated.
+// This ensures to have always at least a usable image left.
+func (rs *RethinkStore) DeleteOrphanImages(images *metal.Images, machines *metal.Machines) (metal.Images, error) {
```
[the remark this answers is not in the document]
majst01: done

cmd/metal-api/internal/datastore/image.go
```diff
@@ -35,3 +47,146 @@ func (rs *RethinkStore) DeleteImage(i *metal.Image) error {
...
+				return nil, fmt.Errorf("unable to delete image:%s err:%v", image.ID, err)
+			}
+			result = append(result, image)
+		}
+	}
+	return result, nil
+}
+
+// isOrphanImage check if a image is not allocated and older than allowed.
+func isOrphanImage(image metal.Image, machines *metal.Machines) bool {
```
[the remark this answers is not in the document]
majst01: done

cmd/metal-api/internal/datastore/image.go
```diff
@@ -35,3 +47,146 @@ func (rs *RethinkStore) DeleteImage(i *metal.Image) error {
...
+	if time.Since(image.ExpirationDate) < 0 {
+		return false
+	}
+	orphan := true
+	for _, m := range *machines {
+		if m.Allocation == nil {
+			continue
+		}
+		if image.ID == m.Allocation.ImageID {
+			orphan = false
```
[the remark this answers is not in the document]
majst01: done

cmd/metal-api/internal/datastore/image.go
```diff
@@ -35,3 +47,146 @@ func (rs *RethinkStore) DeleteImage(i *metal.Image) error {
...
+	}
+
+	matcher := "~"
+	// if patch is given return a exact match
+	if sv.Patch() > 0 {
+		matcher = "="
+	}
+	constraint, err := semver.NewConstraint(matcher + sv.String())
+	if err != nil {
+		return nil, fmt.Errorf("could not create constraint of image version:%s err:%v", sv, err)
```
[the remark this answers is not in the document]
majst01: done

cmd/metal-api/internal/metal/image.go
majst01: removed

cmd/metal-api/internal/metal/image.go
majst01: good point, done

cmd/metal-api/internal/service/image-service.go
majst01: good point, done

majst01:
Actual implementation would require to add ExpirationDate to every image, can by done with the deployment, but is not automatically modified from metal-api. I think this is good enough. WDYT?

cmd/metal-api/internal/datastore/image.go
```diff
@@ -35,3 +47,146 @@ func (rs *RethinkStore) DeleteImage(i *metal.Image) error {
...
+	}
+	if latestImage != nil {
+		return latestImage, nil
+	}
+	return nil, fmt.Errorf("no image for os:%s version:%s found", os, sv)
+}
+
+// GetOsAndSemver parses a imageID to OS and Semver, or returns an error
+func (rs *RethinkStore) GetOsAndSemver(id string) (string, *semver.Version, error) {
+	imageParts := strings.Split(id, "-")
```
ulrichSchreiner, line 160: only cosmetic, but i think a `strings.SplitN(id, "-", 2)` would work too and you do not need to join the version-parts later.

majst01: pointed out a inconsequence and missing test, modified and clarified with test

majst01:
One problem remains:
Migration of current machine.Allocation has a chicken-egg problem, migrating to new images is only possible if new images already configured. We need to discuss howto solve this issue.

ulrichSchreiner:
the only *problem* with the `migrate` endpoint is, what we want to do in the future with this endpoint.

 - leave it as it is for documentation purposes
 - use it in the future for different types of migrations and remove the implementation in the future to have an empty endpoint. so the desription of the endpoint (and the name of the method) should not be `migrateImages`
 - use it in the future for different types of migrations but only ADD new migrations. so the desription of the endpoint (and the name of the method) should not be `migrateImages` but the implementation should contain the current code (and more to come)

if we want to support migrations in the future, shouldn't we think about some sort of a schema-version table in the database, so we can create new schema's and do migrations from old-schema to new-schema?

last but not least: as we discussed i am pretty unsure if it wouldn't be better to simply write a migration program and execute it manually against the database, as the whole thing here is a one-shot action which is only needed in environments with old data structure. i'm not sure how many installations there are in the wild which need this.

cmd/metal-api/internal/datastore/image.go
```diff
@@ -35,3 +47,210 @@ func (rs *RethinkStore) DeleteImage(i *metal.Image) error {
 func (rs *RethinkStore) UpdateImage(oldImage *metal.Image, newImage *metal.Image) error {
 	return rs.updateEntity(rs.imageTable(), newImage, oldImage)
 }
+
+// MigrateMachineImages check images of all machine allocations and migrite them to semver images
+// must be executed only once.
```
Gerrit91, line 52: Wouldn't this be a typical task for a database migration?

majst01: sure but how ?

cmd/metal-api/internal/datastore/image.go
```diff
@@ -35,3 +47,210 @@ func (rs *RethinkStore) DeleteImage(i *metal.Image) error {
 func (rs *RethinkStore) UpdateImage(oldImage *metal.Image, newImage *metal.Image) error {
 	return rs.updateEntity(rs.imageTable(), newImage, oldImage)
 }
+
+// MigrateMachineImages check images of all machine allocations and migrite them to semver images
```
Gerrit91, line 51: Typo: `migrite`

cmd/metal-api/internal/datastore/image.go
Gerrit91: Mh... there are many projects that help with it. I already built his in at some point in time but decided it was too early to add it to the code base. But I think by now we should consider adding such a mechanism as otherwise we clutter the codebase with migration functions.

cmd/metal-api/internal/metal/image.go
```diff
@@ -3,15 +3,63 @@ package metal
...
+	"time"
 )
 
 // An Image describes an image which could be used for provisioning.
 type Image struct {
 	Base
 	URL      string                    `rethinkdb:"url" json:"url"`
 	Features map[ImageFeatureType]bool `rethinkdb:"features" json:"features"`
+	// OS is parsed from id and is the first part, specifies operating system derivate, internal usage only
+	OS string `rethinkdb:"os" json:"os"`
```
Gerrit91, line 15: Unrelated but I am confused that we have `json` annotations here again... I remember that I removed all of them at some point in time. Not sure why and when they came back.

cmd/metal-api/internal/datastore/image.go
```diff
@@ -35,3 +47,210 @@ func (rs *RethinkStore) DeleteImage(i *metal.Image) error {
...
+// must be executed only once.
+func (rs *RethinkStore) MigrateMachineImages(machines metal.Machines) (metal.Machines, error) {
```
Gerrit91, line 53: Why did you make a machines parameter and not just always get all the machines? Is there a specific reason?

majst01: this helps during tests

cmd/metal-api/internal/datastore/image_test.go
```diff
@@ -217,3 +221,305 @@ func TestRethinkStore_UpdateImage(t *testing.T) {
...
+		name    string
+		id      string
+		images  []metal.Image
+		want    *metal.Image
+		wantErr bool
+	}{
+		{
+			name:    "simple",
+			id:      "ubuntu-19.04",
+			images:  []metal.Image{i10, i7, i3, i8, i6, i1, i9, i5, i2, i4, i11},
```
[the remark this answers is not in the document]
majst01: good point, will change

majst01:
tested:
- create new semver images
- run migrate
- deleted legacy images
- created/destroyed machine with metalctl with specific image and image without patch version
- created kubernetes cluster without modified cloudprofile

Issue #22, opened 2020-03-13
Replace long-polling with websocket


majst01:
The current implementation uses long-polling for the `wait` call. A machine calls this function and *hangs* until it is allocated by another client. There is a timeout when the server closes the hanging call (to prevent other network timeouts) and the client will instantly call `wait` again.

When such a timeout occurs, the server returns a 504 status code (Gateway Timeout) which itself is ok. But there are problems with this approach:

 - the log is full of 504 errors which are not really errors. 
 - if we later want to monitor the log for 5xx errors, we have to filter.
 - if we later want to monitor the response-durations, we always have very long response times for the `wait` call, so we also have to filter here.

I'd propose an implementation with websockets, where the client calls the `wait` endpoint which is not a rest-function but the endpoint to a websocket which stays open until one of the partners (client or server) closes the connection. If the connection is lost although it wasn't closed we log an error because this is really an error. When a machine is allocated the machine-struct will be transmitted through the websocket to the client. 

The endpoint `wait` should not be implemented with the restful framework, perhaps it should use a complete different rootpath in the URL with a specific handler in the code. In this case we can exclude this call very easily from generating some metrics like `...response_duration` as this makes no sense with websockets.

### Implications

When using websockets we have to deal with reverse proxies in front of the API service. As websockets are implemented with the `Connection` header which itself is a Hop-By-Hop header we have to configure the reverse proxy to forward the websocket to the metal-api.

---

Sounds good and the way to go, i do not know websockets as deep as you do, but is it still possible to do this wait call through metal-core as we do now ? Or do we need to change the way metal-hammer talks to that endpoint in a more general way ?

---

metal-core has to use `httputil.ReverseProxy` to provide an endpoint for metal-hammer and proxy it to metal-api. go1.12 implemented [this feature](https://go-review.googlesource.com/c/go/+/146437/), i don't see any big problems here.

---

majst01:
@kolsa maybe not exact problem description but can be used as the issue for "use grpc for waitforallocation". Please update here accordingly.

kolsa:
The wait-for-allocation feature is now implemented using gRPC. These are the relevant PRs:

https://github.com/metal-stack/metal-roles/pull/5
https://github.com/metal-stack/mini-lab/pull/7
https://github.com/metal-stack/metal-hammer/pull/14
https://github.com/metal-stack/metal-core/pull/14
https://github.com/metal-stack/metal-api/pull/55
[...]

PR #55, opened 2020-04-20
use grpc server regarding wait-for-allocation feature

kolsa:


cmd/metal-api/internal/grpc/waitServer.go
```diff
@@ -0,0 +1,68 @@
...
+				// Close stream regularly
+				return nil
+			}
+
+			timeout := time.Now().Add(5 * time.Second)
+			ticker := time.NewTicker(500 * time.Millisecond)
+			for now := range ticker.C {
+				m, err := s.ds.FindMachineByID(id)
+				if err != nil {
+					return err
```
ulrichSchreiner, line 59: if the error is returned, the ticker is not stopped

cmd/metal-api/internal/grpc/waitServer.go
```diff
@@ -0,0 +1,68 @@
...
+				return err
+			}
+
+			if allocated {
+				// Close stream regularly
+				return nil
+			}
+
+			timeout := time.Now().Add(5 * time.Second)
+			ticker := time.NewTicker(500 * time.Millisecond)
```
ulrichSchreiner, line 55: im unsure if this is a good idea. this ticker queries the machine table every 0.5secs (2 times per sec). ok, after 5secs there is a timeout, but the outer loop still executes this code again.
if we have 100 machines waiting, we will have ~200 db-queries/sec to find the allocated machine. i don't know if this is a problem, but the whole thing does not looks like it scales with many machines ...

majst01: Soon we have 300machines

Gerrit91:
References #22-

Gerrit91:
@kolsa We merged #66 and you will now need to rebase. After that, we can start testing your changes thoroughly.

Gerrit91:
As far as I can see, in the current implementation, a random machine candidate is being picked from the machine table. However, I don't see that this respects whether this machine is actually "ready" for being allocated, meaning that this machine could also have booted in the UEFI shell. In this case a user would allocate a machine and nothing is gonna happen?

Side note: The wait table gets not cleaned up in the database, which would require a migration as implemented in a proposal in https://github.com/metal-stack/metal-api/pull/70.

cmd/metal-api/internal/grpc/server.go
```diff
@@ -0,0 +1,81 @@
...
+	"google.golang.org/grpc/keepalive"
+	"io/ioutil"
+	"net"
+	"time"
+)
+
+func Serve(ws *WaitServer) {
+	logger := zapup.MustRootLogger().Sugar()
+
+	grpcPort := viper.GetInt("grpc-port")
```
Gerrit91, line 20: I think, we tried to prevent using viper from outside of the `main` package because it could get confusing when accessing variables from anywhere. Do you think it would be possible to wrap these parameters into a struct and pass this as a parameter?

cmd/metal-api/internal/grpc/server.go
```diff
@@ -0,0 +1,81 @@
...
+	)
+	v1.RegisterWaitServer(grpcServer, ws)
+
+	tlsConfig := &tls.Config{
+		NextProtos: []string{"h2"},
+	}
+	if viper.GetBool("grpc-tls-enabled") {
+		cert, err := ioutil.ReadFile(viper.GetString("grpc-server-cert-file"))
+		if err != nil {
+			logger.Fatalw("failed to serve gRPC", "error", err)
```
Gerrit91, line 43: I think, returning an error would be a little more elegant. We should try to avoid `Fatal` and `panic` in the code as much as possible.

cmd/metal-api/internal/grpc/waitServer.go
```diff
@@ -0,0 +1,138 @@
...
+	bus.Publisher
+	ds        *datastore.RethinkStore
+	logger    *zap.SugaredLogger
+	queueLock *sync.RWMutex
+	queue     map[string]chan bool
+}
+
+func NewWaitServer(ds *datastore.RethinkStore, publisher bus.Publisher) (*WaitServer, error) {
+	tlsCfg := &bus.TLSConfig{
+		CACertFile:     viper.GetString("nsqd-ca-cert-file"),
```
Gerrit91, line 37: Same here about viper variable access.

cmd/metal-api/internal/grpc/waitServer.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+	c, err := bus.NewConsumer(zapup.MustRootLogger(), tlsCfg, viper.GetString("nsqlookupd-http-addr"))
+	if err != nil {
+		return nil, err
+	}
+
+	s := &WaitServer{
+		Publisher: publisher,
+		ds:        ds,
+		logger:    zapup.MustRootLogger().Sugar(),
```
Gerrit91, line 48: I think, I would prefer passing the logger as a parameter such that the caller can decide which kind of relationship the logger has to the root logger. Also, I would like to get rid off our zapup package because it does have extra code that is memory expensive.

kolsa:
> As far as I can see, in the current implementation, a random machine candidate is being picked from the machine table. However, I don't see that this respects whether this machine is actually "ready" for being allocated, meaning that this machine could also have booted in the UEFI shell. In this case a user would allocate a machine and nothing is gonna happen?
> 
> Side note: The wait table gets not cleaned up in the database, which would require a migration as implemented in a proposal in #70.

Good catch! Since the wait table is not part of the implementation any more I would suggest to update the machine status to a new `WaitingState` state whenever a machine is registered and use this state as filter criterium when selecting a machine candidate rather than `AvailableState`

cmd/metal-api/internal/service/firewall-service.go
```diff
@@ -17,7 +17,7 @@ import (
 	v1 "github.com/metal-stack/metal-api/cmd/metal-api/internal/service/v1"
 	"github.com/metal-stack/metal-api/cmd/metal-api/internal/utils"
 
-	restfulspec "github.com/emicklei/go-restful-openapi/v2"
+	restfulspec "github.com/emicklei/go-restful-openapi"
```
majst01, line 20: I think this is wrong

PR #58, opened 2020-04-23
Persist sync status of switches

### LINKED ISSUE #50: Show last sync time with metal-core on switch ls
I think, it would be a cool addition if we could show the last time when metal-core synced it's network config with the metal-api.

mwindower:
Closes #50

### DIFF
--- a/cmd/metal-api/internal/service/switch-service.go
+++ b/cmd/metal-api/internal/service/switch-service.go
@@ -3,6 +3,7 @@ package service
 import (
 	"fmt"
 	"net/http"
+	"time"
 
 	restful "github.com/emicklei/go-restful"
 	restfulspec "github.com/emicklei/go-restful-openapi"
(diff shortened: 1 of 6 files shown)


mwennrich:
idea for a follow-up issue: add a metric like metal_api_switch_sync_errors_total or something, so that we can alert on high error rates

cmd/metal-api/internal/service/switch-service.go
```diff
@@ -125,6 +136,46 @@ func (r switchResource) deleteSwitch(request *restful.Request, response *restful
...
+	}
+
+	if requestPayload.Error == nil {
+		s.LastSync = sync
+	} else {
+		sync.Error = requestPayload.Error
+		s.LastSyncError = sync
+	}
+
+	err = r.ds.UpdateSwitch(&old, s)
```
majst01, line 167: This might fail because we have a quite high update frequency on the switch entity, but i think this is ok.

cmd/metal-api/internal/service/v1/switch.go
```diff
@@ -54,12 +55,19 @@ type SwitchRegisterRequest struct {
 	SwitchBase
 }
 
+type SwitchNotifyRequest struct {
+	Duration time.Duration `json:"sync_duration" description:"the duration of the switch synchronization"`
```
majst01, line 59: Good idea to have the sync duration stored, but how can we access it ? Switchresponse does not contain it.
Maybe late we can report this in the metrics ?

mwindower: We can report this as a metric later but the SwitchResponse was enriched with additional fields as well.

majst01: sure but not he duration, as  far as i can see

cmd/metal-api/internal/service/v1/switch.go
```diff
@@ -81,9 +89,11 @@ func NewSwitchResponse(s *metal.Switch, p *metal.Partition, nics SwitchNics, con
 		SwitchBase: SwitchBase{
 			RackID: s.RackID,
 		},
-		Nics:        nics,
-		Partition:   *NewPartitionResponse(p),
-		Connections: cons,
+		Nics:          nics,
+		Partition:     *NewPartitionResponse(p),
+		Connections:   cons,
+		LastSync:      s.LastSync,
```
mwindower, line 95: @majst01 here the complete sync struct is passed to the SwitchResponse

majst01: stupid me

PR #63, opened 2020-04-30
machine create selects wrong image in case of firewalls

majst01:
compare image os instead of same prefix of the os the return the most recent image


Gerrit91:
Tried out in test control plane and is working.

PR #66, opened 2020-05-04
Free machine async

ulrichSchreiner:
implement async deletion of machines and connected ressources


cmd/metal-api/internal/metal/switch.go
```diff
@@ -40,9 +40,7 @@ type SwitchSync struct {
 // which should not happen in a real environment.
 func (s *Switch) ConnectMachine(machine *Machine) {
 	// first remove all existing connections to this machine.
-	if _, has := s.MachineConnections[machine.ID]; has {
```
majst01, line 43: I have found this issue in previous checks, glad to see this fixed

cmd/metal-api/internal/service/async-actor.go
```diff
@@ -0,0 +1,167 @@
...
+
+	err = publishDeleteEvent(pub, m, a.Logger)
+	if err != nil {
+		return err
+	}
+
+	// call the releaser async
+	err = a.machineReleaser(m)
+	if err != nil {
+		// log error, but what should we do here? we already called
```
Gerrit91, line 59: Can't we put `deleteVRFSwitches`, `publishDeleteEvent`, `machineUpdate` all into the machineReleaser and make this idempotent? Such that the api tries to accomplish all these things again and again until success? Does this have any bad side effects?

cmd/metal-api/internal/service/async-actor.go
```diff
@@ -0,0 +1,167 @@
...
+			}
+
+			// at this point the machine is removed from the IP, so the release of the
+			// IP must not fail, because this loop will never come to this point again because the
+			// begin of this loop checks if the IP contains the machine.
+			// so we fork a new job to delete the IP. if this fails .... well then ("houston we have a problem")
+			// we do not report the error to the caller, because this whole function cannot be re-do'ed.
+			a.Infow("async release IP", "ip", *ip)
+			if err := a.ipReleaser(*ip); err != nil {
+				// what should we do here? this error shows a problem with the nsq-bus system
```
Gerrit91, line 116: This is probably a case for our "garbage collector" or IPAM syncer or whatever we want to call that. An algorithm that finds such IP addresses that belong to "nothing" and cleans them up. I also do not know how to deal with this more adequately than that.

cmd/metal-api/internal/service/ip-service.go
```diff
@@ -28,17 +30,23 @@ type ipResource struct {
...
 		webResource: webResource{
 			ds: ds,
 		},
 		ipamer: ipamer,
 		mdc:    mdc,
 	}
+	var err error
+	ir.actor, err = newAsyncActor(zapup.MustRootLogger(), ep, ds, ipamer)
+	if err != nil {
+		panic(err)
```
Gerrit91, line 48: Does this still work for creating the swagger spec through the CLI command?
Maybe also returning an error would be a little friendlier.

cmd/metal-api/internal/service/async-actor.go
```diff
@@ -0,0 +1,167 @@
...
+	if err == nil {
+		// if someone calls freeMachine for the same machine multiple times at the same
+		// moment it can happen that we already deleted and released the IP in the ipam
+		// so make sure that this IP is not already connected to a new machine
+		if len(dbip.GetMachineIds()) > 0 {
+			a.Info("do not delete IP, it is connected to a machine", zap.Any("ip", ip))
+			return nil
+		}
+		s := dbip.GetScope()
+		if s != metal.ScopeProject && ip.Type == metal.Static {
```
Gerrit91, line 143: I don't not really get this condition. The log message says something that this condition does not express. I am not sure what should actually be prevented here.

Can there even be static IPs on another scope than Project Scope?

cmd/metal-api/internal/service/ip-service.go
```diff
@@ -187,36 +195,17 @@ func (ir ipResource) freeIP(request *restful.Request, response *restful.Response
...
-		return
-	}
 	err = response.WriteHeaderAndEntity(http.StatusOK, v1.NewIPResponse(ip))
 	if err != nil {
 		zapup.MustRootLogger().Error("Failed to send response", zap.Error(err))
 		return
 	}
 }
 
-func validateIPDelete(ip *metal.IP) error {
```
Gerrit91, line 212: I think the safer variant would be to leave this validation in the web service and not move it to the ip releaser. But maybe this is just because I do not know what this validation is actually meant for.

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -117,16 +117,24 @@ type Allocator func(Allocation) error
...
 		},
-		Publisher: pub,
+		publisher: pub,
 		ipamer:    ipamer,
 		mdc:       mdc,
 	}
+	var err error
+	r.actor, err = newAsyncActor(zapup.MustRootLogger(), ep, ds, ipamer)
+	if err != nil {
+		panic(err)
```
Gerrit91, line 135: Same question here regarding the swagger command and `panic` as above.

cmd/metal-api/internal/service/async-actor.go
ulrichSchreiner: the `publishDeleteEvent` must not be retried if anything fails. if in a later (async) step the release of an ip address fails the machine will already be released. so it is possible that the machine is already allocated by another one. so we must not publish the delete event again.

the `deleteVRFSwitches` is another part. i'm not quite deep in the code but i think this has to be done before the machine is deleted.

cmd/metal-api/internal/service/async-actor.go
ulrichSchreiner: i only moved the code from here: https://github.com/metal-stack/metal-api/blob/4a2153efe97d37ebffbc8b419b4556c08095b3b8/cmd/metal-api/internal/service/ip-service.go#L183

to the actor. about two weeks ago we discussed this validation logic, because i also was a little puzzled about this one. it was created by @mwindower in this commit https://github.com/metal-stack/metal-api/commit/472e4aa1a3f014068ddb914185ed11a9a1bcc4ae

cmd/metal-api/internal/service/async-actor.go
Gerrit91: I still do not fully get why we do not put `UpdateMachine` into the reconcile. Did you put it there because you wanted to make a clear statement for the consumer that after this endpoint was called it is guaranteed that this machine has no allocation anymore?  

If the call fails there, the user needs to call `machineFree` again but there are already reconcile loops working on the deletion. I don't know if it is good to motivate the user calling this endpoint over and over again. Also, wouldn't it be a little more robust to put `UpdateMachine` in the reconcile to make sure that networks were cleaned up properly before allowing anyone else to allocate this machine again? I think, at most providers, when you call the deletion of a machine, you will continue to see it until it was fully removed by the backend.

And if we put `UpdateMachine` in the loop, we also need to put `deleteVRFSwitches` and `publishDeleteEvent` in the loop because otherwise the machine could already have booted into the old OS (hanging in `Phoned Home`). So we need to call both functions right before `UpdateMachine`. `UpdateMachine` needs to come last because after that another user can allocate the machine again, which should be the last step to finish off deletion.

cmd/metal-api/internal/service/async-actor.go
Gerrit91: @mwindower Can you please provide your opinion to this? Is this validation required for IP cleanup on `machineFree` as well or only when a user calls IP delete?

mwindower: This is about static IPs that were directly assigned to machines. If the scope is `ScopeMachine` this means that there is at least one machine that uses it. In that case `freeIP` should fail. 
`freeMachine` or the actor should only delete the machine tag from those IPs but never delete it.

cmd/metal-api/internal/service/async-actor.go
ulrichSchreiner: well, imho the `metal-api` should not care about callers. if the caller of the api has a reconcile loop this is a valid strategy; it would also be a strategy to handle errors in another way. 

and the caller has to handle errors. the deletion of a machine **can** fail: we check for a locked machine (perhaps we will have more validations in the future) or we can have technical problems communicating with the DB or NSQ. 

we have to make a balance between the promise to the caller that the deletion really works and some errors which can happen. doin'g everything async will always return an OK to the caller in any case (except when we have a communication error with NSQ).

i think when the `machineFree` returns, the machine should be unallocated. so the caller of this function can be sure that the machine is destroyed/unallocated. the dependent ressources like IP adresses will be cleaned up async.  this guarantees some sort of consistency: after a successfull deletion of a machine the caller can immediatly query the same machine and the api should return an unallocated machine.

the current code does this in a sync way:
  1. remove the machine from the network
  2. send a deletion event to metal-core (triggers a reboot)
  3. updates the allocation state

now the machine should be free; and it also forks an async cleaup-job for all the ressources which are still allocated by the machine.
i'm thinking about switching the order of 2. and 3. because metal-core triggers a reboot, and the machine then queries if it is allocated; this should return nil; but the current code can not guarantee this.

doin'g the unallocation synchron makes sure that the caller can be sure that the machine is free after a successfull call. and i think this is the intention of the caller.

but i see your point too :smiley: 

doin'g anything in a async block would work too. but than we change the semantic of our api. this would be the first endpoint which can be called successfully but if you query the api direct after the `free` it still returns the allocation for this machine. so a fully async implementation adds some form of inconsistency to our api. i'm not sure if we want this (here).

in this case the caller would need some form of a promise (a callback) which will be called after the machine is really free. or the caller polls until the machine is really unallocated. not every client of this api is a machine reconciler which does not care about the machine any more ...

Gerrit91: Cool, thanks for your detailed explanation. It now makes more sense to me that this is the right approach for the time being.

If at some point in the future we still observe a need for more robustness, I think that we could discuss again if it wouldn't be better to put as much into the reconciliation as possible (which probably feels a bit more like dealing with "kubernetes"). I can imagine that more tasks in the reconciliation can give us more control over our workflows and the final state can be reached more reliably without getting into states of inconsistent data.

cmd/metal-api/internal/service/async-actor.go
Gerrit91: Ok. I think this information helps me a bit. Let's go through two scenarios:

Scenario 1:

- A user allocates a static IP address
- A user allocates a machine and adds the IP via `--ips` 
- A user calls machine free 
  - The IP should not be released 
  - The machine tag should be removed from the IP

Scenario 2 (I hope this is correct):

- A user allocates a static IP address
- A user allocates a machine and adds the IP via `--ips` 
- A user calls ip free on the IP address
  - The IP should not be released 
  - The IP machine tag should not be removed because there is still a machine depending on it (otherwise we have a machine with an IP in the machine allocation that the machine does not own anymore)
  - This actually means that nothing happened

Therefore I guess that we should actually check if a machine still depends on an IP address in `freeIP` before handing it over to the reconcile. Just raise an error and let the user free the machines first before he can free the static ip. With the current implementation it would look to the user like `freeIP` was a success but actually nothing happened except our logger said "do not delete static IP" but unfortunately the user still owns the IP without knowing why. 

I am not sure if I made a mistake in my reasoning though. But if this is correct, the check does not belong in the reconcile but only in `freeIP`.

Gerrit91: When I think about this, I would assume that it even does not matter in this case if an IP is static or ephemeral. The error message should be something like "cannot free ip address with machines depending on it". And it should always be checked in `freeIP`.

mwindower: Yes, I agree:
- always check for machines depending on an IP (ephemeral as well as static in `freeIP` and `freeMachine`)
- let `freeMachine` only delete ephemeral IPs

ulrichSchreiner: @mwindower is the condition 
~~~
if s != metal.ScopeProject && ip.Type == metal.Static {
~~~
identical to
~~~
if s == metal.ScopeMachine
~~~
?

this would be much more intuitive.

cmd/metal-api/internal/service/ip-service.go
ulrichSchreiner: You're right. the old implementation returned an error the the caller if the validation failed. My new implementation always returns OK but in the async code the validation could fail. i'll change this

cmd/metal-api/internal/service/async-actor.go
mwindower: As @Gerrit91 mentioned: the whole condition is unnecessary. We could just remove the whole if-block `if s != metal.ScopeProject && ip.Type == metal.Static`

Gerrit91:
Tried it out in our test control plane and can say that ip allocate / free, machine create / remove and cluster create / remove is working. Deployment needs to add `METAL_API_NSQLOOKUPD_ADDR` environment variable for the metal-api.

PR #68, opened 2020-05-07
all other states are reported as such

majst01:


cmd/metal-api/internal/service/v1/partition.go
```diff
@@ -45,6 +45,7 @@ type ServerCapacity struct {
 	Free      int    `json:"free" description:"free servers with this size"`
 	Allocated int    `json:"allocated" description:"allocated servers with this size"`
 	Faulty    int    `json:"faulty" description:"servers with issues with this size"`
+	Other     int    `json:"other" description:"servers neither free,allocated or faulty with this size"`
```
Gerrit91, line 48: missing whitespace between free and allocated

majst01: ok

Issue #69, opened 2020-05-13
Provide Mechanism for Database Migration


Gerrit91:
We will come into situations where we need to migrate our database in order to support upgrades from older versions.

The solution should be simple and stable.

Gerrit91:
The mechanism should be part of the "init-db" command, which we use in our deployments anyway before we rollout a new update. 

A solution where you can define the migrations directly in Go would be preferable, I guess. Something simple like this but for RethinkDB would be nice:

https://github.com/go-pg/migrations

majst01:
in masterdata-api we already use: 

github.com/lopezator/migrator

Gerrit91:
Also looking good but there does not seem to be a tool for rethinkdb which does the same kind of thing. Except for this project https://github.com/Coccodrillo/rethinkdb-migrate, which does not seem to be very active anymore...

majst01:
this is a dead project, maybe we could add rethinkdb support to the migrator i mentioned, i would really try to avoid using multiple mechanisms, especially because we use pg from metal-api as well.

PR #70, opened 2020-05-14
Provide allocation UUIDs for IP addresses.

Gerrit91:
This is required for accounting as otherwise we will not be able to distinguish if an IP address was re-allocated or not.

The database requires a migration for this as we started to discuss in #69 because the allocation UUID needs to be generated for all "old" IP address entities in the database.


cmd/metal-api/internal/service/machine-service.go
```diff
@@ -1311,12 +1312,18 @@ func gatherUnderlayNetwork(ds *datastore.RethinkStore, allocationSpec *machineAl
...
+			return nil, fmt.Errorf("unable to create uuid for IP allocation: %v", err)
+		}
+
 		ipAddress, ipParentCidr, err := allocateIP(n.network, "", ipamer)
 		if err != nil {
 			return nil, fmt.Errorf("unable to allocate an ip in network: %s %#v", n.network.ID, err)
 		}
 		ip := &metal.IP{
 			IPAddress:        ipAddress,
+			AllocationUUID:   allocationUUID.String(),
```
majst01, line 1326: This is not the only place where a IP is allocated, adding the allocationUUID must be done in a central place

Gerrit91: Mh, yes, that would be better. But it's also a bigger refactoring to make this look adequate... I am not sure, if this refactoring should be contained in this PR.

The IP create function of the IP service and the machine network auto ip allocation should be able to use a common `ipCreate` function that hands over an IPSpec as a parameter, I guess.

majst01: TBH, this PR looks crazy as hell

Gerrit91:
I wrote down a couple of lines that would enable us to run zero-downtime migrations of the RethinkDB sufficient for most of the update scenarios we did encounter in the past. I did it this way because there are no existing libraries available for RethinkDB, which would do such a thing and providing such a mechanism is a little different for RethinkDB than it is for usual SQL databases. You need to keep in mind that RethinkDB does not have a schema and that the model is actually described by the database clients, there are no transactions or things like that and we have multiple instances of clients running at the same time.

Therefore, the idea was that the deployment runs a post-upgrade hook that migrates the database to the latest state after new pods were rolled out. Reason that it is run after the deployment is that old clients should not be writing old models into the database after the migration has run. Of course there is one downside of running the migration after the rollout, which is that for a short period of time a client will not return the migrated data from the database, which is "kind-of" acceptable considering that the consumers of the metal-api need to update to the new version in most scenarios.

Summary of the approach:
- Add a "migration" table that holds the current migration version of the database
- A `migrate` command triggers the migration and does a migration when a newer version is available
- The pods of our metal-api replica set do not use the admin user anymore, such that the database can be put into read-only mode (this helps to avoid possible modification / concurrency issues during migration tasks)

Nevertheless, there can be database changes that require a downtime (e.g. changing a type of an existing field). In this case we would need to cut off client connections if we do not want to return with internal crashes to the clients until the migration has run... this would need to be done manually or require more sophisticated deployment strategies, which can determine if downtime is necessary and scale down / up the metal-api replica set for the migration. At the moment, we are not in this situation that we need such an upgrade path and I think that most of the time we can get away with this simple mechanism instead. It will certainly require some elaboration when we come into such a situation.

Maybe we should also put the "SemVer" migration that we had in the past into the list of migrations. Then we would support updating from metal-api v0.6.x to v0.7.x and later, which is not the case at hte moent and could lead to weird issues for someone who attempts such a version upgrade.

Gerrit91:
Of course I am open for other ideas. This actually did not took so much time to implement (few hours). If there are better ideas, please tell me. It's just that I do not want to put in a "temporary migration endpoint" again for this because I think that breaks the update experience.

cmd/metal-api/internal/service/machine-service.go
Gerrit91: I can't deny that. But it looked simpler to me than other approaches in regards of time and suitability. I am open for ideas and comments. :)

Gerrit91:
@ulrichSchreiner We already talked about this a little in the background. What do you think about this? Would this be mergeable from your point of you or is it too "crazy as hell"?

majst01:
First question: how can read-only replicas ever gets back into read-write mode after the upgrade was done ?

Gerrit91:
It's done dynamically via the "grant" command of rethinkDB: https://rethinkdb.com/api/javascript/grant

The read-only mode should only be activated for a very short amount of time, just as long as the migration takes.

majst01:
Probably replaced partly with: https://github.com/metal-stack/go-ipam/pull/28

Gerrit91:
As this PR is likely not to be merged because the IP UUIDs will probably become part of a newer go-ipam version, I am closing this in favor of #98.

PR #81, opened 2020-07-01
Corrections on the api spec.

Gerrit91:
Prior version has semantic errors (checked via [editor.swagger.io](https://editor.swagger.io/)).

Clients of some languages (e.g. Python) generate request validations, which enforces to define some parameters with the current spec even though they should be optional (e.g. search and update requests). Without these changes, these clients are quite unusable.


Gerrit91:
References https://github.com/metal-stack/metal-python/commit/1cdea7b8905fa3c787046c8e0d8d3dc671c2967a.

cmd/metal-api/main.go
```diff
@@ -600,11 +600,11 @@ func enrichSwaggerObject(swo *spec.Swagger) {
...
 			License: &spec.License{
 				LicenseProps: spec.LicenseProps{
-					Name: "MIT",
-					URL:  "http://mit.org",
+					Name: "AGPL-3.0",
+					URL:  "https://www.gnu.org/licenses/agpl-3.0.de.html",
 				},
 			},
-			Version: "1.0.0",
+			Version: v.Version,
```
Gerrit91, line 607: This does not work unfortunately because we call `make spec` manually and push the result then. The version tag is incorrect at this point in time though... don't know how we can handle this more adequately. :(

cmd/metal-api/internal/service/v1/ip.go
```diff
@@ -41,6 +41,10 @@ type IPResponse struct {
 }
 
 func NewIPResponse(ip *metal.IP) *IPResponse {
+	tags := ip.Tags
```
majst01, line 44: ip might be nil !

Gerrit91: Mh... yes, in theory this could happen. We do not check this in all these kind of mapping functions and just access the IP entity without any precautions. It would also be weird if it would be `nil`, but if you want I'll add a verification to all these functions?

Gerrit91:
This must be a next minor version release because it could break existing go-clients. After this, metal-go needs to be fixed and all clients that use the next api version must update to the latest metal-api version.

cmd/metal-api/main.go
Gerrit91: We decided on just not putting the version in there.

majst01:
We really have to think twice how we move this into production because all requests to find{machine,network,ip,image} are now different, they have the arguments optional which is correct, but will cause existing calls from already installed applications in all clusters to fail, most prominent:
- metal-ccm
- gepm
- mcm

This have the potential that the existing clusters fall apart, because for example the network connectivity to the service does not work anymore because metal-ccm is not able anymore to query ips and configure metallb with a empty ip list.

Gerrit91:
Yes, I agree, we need to be careful and really have to update all the clients at once. 

However, if we can't do such a change anymore, I think we did something wrong. We should never be completely stuck with anything. :confused:

majst01:
I don´t say we should not do it, right the opposite, but we must ensure nothing breaks, and changing all binaries at all is no enough. because once the api is deployed with this change, all existing clusters may already be broken, or go into a broken state in a way that updating ccm for example does not even work anymore.

Gerrit91:
We could theoretically introduce a `v2` API and put all changed requests in there. We would then need to introduce functions that map the requests to common structs and pass them on to the actual functions in our sources. We could then offer both API versions such that clients have time to migrate. go-restful also allows us to deprecate certain functions. Don't know if that's worth the effort in this early stage of the project. Do we really know if declaring a field optional breaks existing clients?

majst01:
No i think its not worth the effort, we should test if clients break in test, albeit this is quite an effort.

Gerrit91:
I tried out the metal-api build from this pull request in our test environment. 

This PR mainly includes changes that declare optional parameters in search and update requests. With metalctl I tried modified endpoints with and without search filters and did not see any changes in behavior. I assume this is really only a thing for swagger clients that generate validations.

Then, I tried cluster creation / deletion, which also was a success.

The impact of the changes on existing Golang clients seems to be fairly small.

Gerrit91:
Ping?

majst01 (approved):

majst01:
metal-go needs to be adapted now

PR #84, opened 2020-07-14
Use own struct for project and friends

majst01:
Use own struct for project and friends to avoid stackoverflow while creating spec which contains proto project structs

Compiles but untested.
This solves the stackoverflow but is ugly as hell, same has to be done for cloud-api as well which will duplicate the effort also.
This is also a nightmare for maintainability.

Thoughts ?


majst01:
Closing, structs already moved to masterdata-api

PR #93, opened 2020-08-05
Project create/update/delete endpoints

majst01:
Add project create/update/delete enpoints for admins only.


cmd/metal-api/internal/service/project-service.go
```diff
@@ -161,34 +174,221 @@ func (r projectResource) findProjects(request *restful.Request, response *restfu
...
+		pips := result[ip.ProjectID]
+		pips = append(pips, ip)
+		result[ip.ProjectID] = pips
+	}
+	return result, nil
+}
+
+func (r projectResource) machinesByProject() (map[string]metal.Machines, error) {
+	// FIXME implement ScopedMachineFind
+	ms, err := r.ds.ListMachines()
```
Gerrit91, line 377: Why not using `r.ds.SearchMachines` and query for project id?

cmd/metal-api/internal/service/project-service.go
```diff
@@ -161,34 +174,221 @@ func (r projectResource) findProjects(request *restful.Request, response *restfu
...
+	if err != nil {
+		return nil, err
+	}
+	result := make(map[string]metal.Machines)
```
Gerrit91, line 381: There actually is `ByProjectID()` on the `Machines` struct, which is the result of `SearchMachines`

cmd/metal-api/internal/service/project-service.go
```diff
@@ -161,34 +174,221 @@ func (r projectResource) findProjects(request *restful.Request, response *restfu
...
+		Project: *v1p,
+	})
+	if err != nil {
+		zapup.MustRootLogger().Error("Failed to send response", zap.Error(err))
+		return
+	}
+}
+
+func (r projectResource) ipsByProject() (map[string]metal.IPs, error) {
+	ips, err := r.ds.ListIPs()
```
Gerrit91, line 359: Same as for machines (also has `ByProjectID()`)

cmd/metal-api/internal/service/project-service.go
```diff
@@ -161,34 +174,221 @@ func (r projectResource) findProjects(request *restful.Request, response *restfu
...
+
+	existingProject, err := r.mdc.Project().Get(context.Background(), &mdmv1.ProjectGetRequest{Id: requestPayload.Project.Meta.Id})
+	if checkError(request, response, utils.CurrentFuncName(), err) {
+		return
+	}
+
+	// new data
+	projectUpdateData := mapper.ToMdmV1Project(&requestPayload.Project)
+	// created date is not updateable
+	projectUpdateData.Meta.CreatedTime = existingProject.Project.Meta.CreatedTime
```
Gerrit91, line 338: Can't this be validated and return an error when this gets requested? Why do we offer this in the `ProjectUpdateRequest` in the first place?

cmd/metal-api/internal/service/project-service.go
majst01: because then i have to query for every project instead of once and access the machines by Project through the map

cmd/metal-api/internal/service/project-service.go
majst01: same comment as above

cmd/metal-api/internal/service/project-service.go
majst01: Because masterdata-api does not offer seperate Structs for Create and Update for simplicity. But this is of course the downside of this decision

cmd/metal-api/internal/service/project-service.go
majst01: done

cmd/metal-api/internal/service/project-service.go
majst01: use ByProjectID() now

cmd/metal-api/internal/service/project-service.go
majst01: use ByProjectID() now

cmd/metal-api/internal/service/project-service.go
```diff
@@ -73,6 +76,36 @@ func (r projectResource) webService() *restful.WebService {
 		Returns(http.StatusOK, "OK", []v1.ProjectResponse{}).
 		DefaultReturns("Error", httperrors.HTTPErrorResponse{}))
 
+	ws.Route(ws.DELETE("/{id}").
+		To(admin(r.deleteProject)).
```
droid42, line 80: LGTM, a test to prove that the permission checks work would make it perfect

majst01: done

PR #94, opened 2020-08-07
Fail fast

kolsa:


majst01:
requires another change in metal-hammer because the renaming of WaitResponse.

Is this a fix for the spurious missed immediate allocations ?

kolsa:
Yes, this could fix that issue.
Also, see [this PR](https://github.com/metal-stack/metal-hammer/pull/24/files)

cmd/metal-api/internal/grpc/waitServer.go
```diff
@@ -159,14 +162,30 @@ func (s *WaitServer) Wait(req *v1.WaitRequest, srv v1.Wait_WaitServer) error {
...
 				return err
 			}
 		}
 	}
 }
 
+// https://github.com/grpc/grpc-go/issues/1229#issuecomment-302755717
+func sendKeepPatientResponse(srv v1.Wait_WaitServer) error {
+	errChan := make(chan error, 1)
+	ctx := srv.Context()
```
Gerrit91, line 176: Just for me to know... Under which circumstances does this context receive a `Done()`?

kolsa: Whenever the stream to the client completes or breaks

Gerrit91: Okay, then let's do it.

majst01:
But with this change we also need a new metal-hammer as i said, changing proto definitions in place does not retain compatibility !

PR #98, opened 2020-08-19
Introduce rethinkdb migration mechanism.

### LINKED ISSUE #69: Provide Mechanism for Database Migration

Gerrit91:
Closes #69.

Discussion for implementation already started in #70. This is now the extraction of only the database migration mechanism of this PR. Please consider reading the issue, too. Here is the most relevant summary:

> I wrote down a couple of lines that would enable us to run zero-downtime migrations of the RethinkDB sufficient for most of the update scenarios we did encounter in the past. I did it this way because there are no existing libraries available for RethinkDB, which would do such a thing and providing such a mechanism is a little different for RethinkDB than it is for usual SQL databases. You need to keep in mind that RethinkDB does not have a schema and that the model is actually described by the database clients, there are no transactions or things like that and we have multiple instances of clients running at the same time.
> 
> Therefore, the idea was that the deployment runs a post-upgrade hook that migrates the database to the latest state after new pods were rolled out. Reason that it is run after the deployment is that old clients should not be writing old models into the database after the migration has run. Of course there is one downside of running the migration after the rollout, which is that for a short period of time a client will not return the migrated data from the database, which is "kind-of" acceptable considering that the consumers of the metal-api need to update to the new version in most scenarios.
> 
> Summary of the approach:
> 
> Add a "migration" table that holds the current migration version of the database
> A migrate command triggers the migration and does a migration when a newer version is available
> The pods of our metal-api replica set do not use the admin user anymore, such that the database can be put into read-only mode (this helps to avoid possible modification / concurrency issues during migration tasks)
> Nevertheless, there can be database changes that require a downtime (e.g. changing a type of an existing field). In this case we would need to cut off client connections if we do not want to return with internal crashes to the clients until the migration has run... this would need to be done manually or require more sophisticated deployment strategies, which can determine if downtime is necessary and scale down / up the metal-api replica set for the migration. At the moment, we are not in this situation that we need such an upgrade path and I think that most of the time we can get away with this simple mechanism instead. It will certainly require some elaboration when we come into such a situation.
> 
> Maybe we should also put the "SemVer" migration that we had in the past into the list of migrations. Then we would support updating from metal-api v0.6.x to v0.7.x and later, which is not the case at the moment and could lead to weird issues for someone who attempts such a version upgrade.


Gerrit91:
Tested in mini-lab the update from the current develop version to this version, which was looking good.

majst01 (approved):

PR #101, opened 2020-09-11
add bmc superuser password endpoint

kolsa:


cmd/metal-api/internal/grpc/grpcServer.go
```diff
@@ -0,0 +1,191 @@
...
+	FindMachineByID(machineID string) (*metal.Machine, error)
+	UpdateMachine(old, new *metal.Machine) error
+}
+
+type ServerConfig struct {
+	Publisher             bus.Publisher
+	Datasource            Datasource
+	Logger                *zap.SugaredLogger
+	NsqTlsConfig          *bus.TLSConfig
+	NsqlookupdHttpAddress string
```
Gerrit91, line 49: Just from the wording it does not sound odd to me that these things are part of a "server config". 

I would appreciate if we could move service implementations into separate files and these services have their own constructors. At least to me this would make the code much easier to understand from the outside.

Service construction would need to be done from `main.go` in the main init process.

cmd/metal-api/internal/grpc/grpcServer.go
```diff
@@ -0,0 +1,191 @@
+package grpc
```
Gerrit91, line 1: Can we name this file just `server.go`? Otherwise we will introduce "camelCase" for file names together with "hyphenation".

cmd/metal-api/internal/grpc/supwdServer.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"context"
+	"fmt"
+	v1 "github.com/metal-stack/metal-api/pkg/api/v1"
+	"io/ioutil"
+	"strings"
+)
+
+func (s *Server) FetchSupermetalPassword(ctx context.Context, req *v1.SupermetalPasswordRequest) (*v1.SupermetalPasswordResponse, error) {
```
Gerrit91, line 11: `Supermetal` or `SuperMetal`? Or maybe `SuperPassword` is already enough? Why do we need `Metal` in the name after all?

cmd/metal-api/internal/grpc/supwdServer.go
```diff
@@ -0,0 +1,22 @@
...
+	defer ctx.Done()
+
+	bb, err := ioutil.ReadFile(fmt.Sprintf("/supwds/%s.supwd", req.PartitionID))
```
Gerrit91: As HMACs are currently also just set as environment variable, I would do it the same way for now here.

We can create a follow-up issue where we move things like passwords to configuration files, such that a user cannot see the content from watching the pod spec only.

kolsa: I've already prepared a secret and a corresponding /supermetal mountpoint in metal-roles, which is pretty much the same mechanism we already have regarding the grpc certificates.
That being said I would leave this here as is, since it's also the final solution.

majst01:
Generally i am not very happy with this superpassword, this does not reflect the actual use case which is the password for the admin account of the bmc. Would it better to fetch the bmcadmin with username and password ?

cmd/metal-api/internal/grpc/supwdServer.go
kolsa: At least we need some context, since SuperPassword does not say anything regarding the corresponding user, which is the bmc/ipmi superuser (https://github.com/metal-stack/metal-hammer/issues/25).

Since our default bmc/ipmi username is `metal` I chose the name `supermetal` to indicate this fact.

cmd/metal-api/internal/grpc/grpcServer.go
kolsa: Indeed, probably we should agree on either camelCase, hyphonetation-style or under_scoring, since their are all used in distinct metal-stack places.
Personally I pretty much prefer camelCase, since it is shorter, faster to type and even better readable.
Additionally, tests are more clearly identifiable by just looking at the contents of a golang package.

kolsa: Besides this, I would leave the grpc prefix as is, since you can find it easier from within your IDE or via rg.
Also, it doesn't affect the golang code at all.
Nevertheless, the fqn of the server struct itself is in fact `grpc.Server`

cmd/metal-api/internal/grpc/supwdServer.go
Gerrit91: But then we would also need the HMACs to be put into separate files when we want everything to be uniform. This means, that in the end there would be six more files where we do `ioutil.ReadFile`, which I think is not as safe as the existing initialization because we need to ensure that all required parameters are present on startup, otherwise don't even start the API.

Also, if this password can be different for every partition... isn't this actually part of the partition configuration?

cmd/metal-api/internal/grpc/grpcServer.go
kolsa: Good point, readability should always be improved as much as possible

kolsa:
Since the bmcadmin password is publicly known per vendor we do have to provide its password from outside.
Thus I do not see the benefit to fetch the bmcadmin along with its current password.
Why not simply update the password each time metal-hammer runs?
If it's only a naming thing, we already have `metal` as default ipmi user, thus `supermetal` abstracts the name of the bmcadmin, which for example is `USERID` for Lenovo and `ADMIN` for Supermicro

majst01:
agreed but then we also have to disable the vendor provided bmc admin

cmd/metal-api/internal/grpc/supwdServer.go
kolsa: I agree.
Another solution would be to use one bmc admin password for all machines of all vendors.
For this we could leave this implementation as is despite changing the path to something like this `/bmc/admin.pwd`

Gerrit91: I think, if it is just a single file and the existence is ensured at startup, I am fine. If it would be optional (non-breaking for deployments) this would be even better. Only set a password when a user explicitly provides one, otherwise just do not do anything.

kolsa: Since connecting to a machine's bmc is already protected by our mgmt-net I'm also fine using only one password that rules them all.
If we make this password optional, we do not fix the issue of loosing machines (at least from remote access), which happened in the past, if only rarely. Additionally we would have to go the second path, namely to disable the bmc admin user with its publicly-known password and create an new one (`supermetal`) with admin privileges.

Anyways, this second path would have the benefit that an admin user do not care about the machine's vendor in order to access it's bmc, since the admin usernames differ from vendor to vendor.

majst01 (approved):
execution in test before merge required

majst01 (approved):

Issue #102, opened 2020-09-15
Allow attaching shared networks to firewalls


majst01:
It is in some cases required to be able to enable machines from multiple projects have direct access through the firewall to special purpose networks like storage.

The proposed solution is to mark a private machine network as a shared network, which allows all other consumers to attach to this network as well.

/cc @mwennrich @muhittink @Gerrit91

mwindower:
There are different ways to solve this issue.

**1. Using a dedicated network for storage nodes**

Pro
- making storage nodes accessible for clusters is straight forwards 

Contra
- is somehow comparable to a private network but needs manual "allocation" of the prefix and VRF
- machines have no private network at all => validations and ASN determination in the metal-api have to be modified

**2. Using a standard private network without project reference**

Clusters that want storage access would have two private networks
- their "primary" private network for worker IPs
- and the private network for communication with the storage nodes (which has no project reference)

Pro
- is a real private network: prefix, VRF and ASN come without any changes

Contra
- constraints for network allocation have to be weakened
- machine allocation has to differentiate btw. the primary private network and secondary private networks (those without project reference)

After looking into the code the second approach is my favorite one.

WDYT?

majst01:
But howto allocate a "secondary" private network

mwindower:
For the storage nodes we could allocate a private network without project id.
During firewall creation of k8s clusters this network is allowed to be in the list of "additional networks".

majst01:
This way we can modify the creation of the current storage network to have set `private: true` and try again !

majst01:
no we cant, private network is evaluated by looking at a privatesuper given

PR #103, opened 2020-09-15
Multiple private networks

mwindower:
draft implementation to allow secondary private networks for firewalls (that must not be associated to a project)
proposal for #102


mwindower:
To be discussed with @Gerrit91 @majst01

majst01:
Another Approach would be to add a feature to make a private network accessible by others instead of setting projectid to empty string. I could imagine something like:

`metalctl network allocate --shared --network=tenant-super-network --projectid=<id> --partition=<partition> `

This will indicate that firewalls from other projects are allowed to join this network as well.

mwindower:
I added a `shared` flag as proposed, which is only used for private networks.

It can be specified for `network allocate` and `network update`.

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -1100,13 +1101,14 @@ func gatherNetworksFromSpec(ds *datastore.RethinkStore, allocationSpec *machineA
 	// what do we have to prevent:
 	// - user wants to place his machine in a network that does not belong to the project in which the machine is being placed
 	// - user wants a machine with a private network that is not in the partition of the machine
-	// - user wants to define multiple private networks for his machine
-	// - user must define one private network
+	// - user specifies no private network
+	// - user specifies multiple primary private networks (those are not shared)
```
majst01, line 1105: Is this correct, can a machine have multiple primary not shared networks, i doubt not

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -1100,13 +1101,14 @@ func gatherNetworksFromSpec(ds *datastore.RethinkStore, allocationSpec *machineA
...
 	// - user specifies administrative networks, i.e. underlay or privatesuper networks
 	// - user's private network is specified with noauto, which would make the machine have no ip address
 
 	specNetworks := make(map[string]*allocationNetwork)
-	var privateNetwork *allocationNetwork
+	var primaryPrivateNetwork *allocationNetwork
+	var privateNetworks []*allocationNetwork
```
majst01, line 1111: Isn't a shared private network just a network like internet, we need to discuss this

cmd/metal-api/internal/service/machine-service.go
mwindower: Multiple primary networks are not allowed; we should use the same logic for all items of the list

mwindower: This is already the case - nothing to do here

```
        // what do we have to prevent:
	// - user wants to place his machine in a network that does not belong to the project in which the machine is being placed
	// - user wants a machine with a private network that is not in the partition of the machine
	// - user specifies no private network
	// - user specifies multiple primary private networks (those are not shared)
	// - user specifies administrative networks, i.e. underlay or privatesuper networks
	// - user's private network is specified with noauto, which would make the machine have no ip address
```

majst01: got it thanks

majst01:
Some thoughts:
- we should not allow to connect to a private and a shared private network in different partitions
- should we allow machines to allocate ips in a shared network which is not their own network ?
- is the current implementation aware of the fact that a shared network can be the primaryPrivate network in case this is the network they are located ?

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -1155,37 +1155,48 @@ func gatherNetworksFromSpec(ds *datastore.RethinkStore, allocationSpec *machineA
...
 		return nil, fmt.Errorf("the given private network does not belong to the project, which is not allowed")
 	}
 
-	for _, ipString := range allocationSpec.IPs {
-		ip, err := ds.FindIPByID(ipString)
-		if err != nil {
-			return nil, err
+	for _, pn := range privateNetworks {
+		if pn.network.PartitionID != partitionPrivateSuperNetwork.PartitionID {
+			return nil, fmt.Errorf("private networks must be in the partition where the machine is going to be placed")
```
mwindower, line 1168: the constraint "all private networks must be located in the same partition" is enforced here.

mwindower:
> Some thoughts:
> 
> * we should not allow to connect to a private and a shared private network in different partitions
> * should we allow machines to allocate ips in a shared network which is not their own network ?
> * is the current implementation aware of the fact that a shared network can be the primaryPrivate network in case this is the network they are located ?

- first point is already enforced
- in the current implementation we allocate an IP in the shared network, if "a cluster wants to join it" because otherwise we would have to redistribute all participating private networks to the shared network (currently they will have the attribute no-export in the frr config and will not be redistributed); this means every participating firewall will have an IP and do SNAT (which is not that good)
- good point - this is currently not respected!

majst01:
> > Some thoughts:
> > 
> > * we should not allow to connect to a private and a shared private network in different partitions
> > * should we allow machines to allocate ips in a shared network which is not their own network ?
> > * is the current implementation aware of the fact that a shared network can be the primaryPrivate network in case this is the network they are located ?
> 
> * first point is already enforced
> * in the current implementation we allocate an IP in the shared network, if "a cluster wants to join it" because otherwise we would have to redistribute all participating private networks to the shared network (currently they will have the attribute no-export in the frr config and will not be redistributed); this means every participating firewall will have an IP and do SNAT (which is not that good)

I understand that the firewall *must* allocate a IP in the shared network he is not primary member of, but for the worker nodes this would not be possible after the cluster was created. So the route distribution for the ips of the machines in the shared network is required?

> * good point - this is currently not respected!

mwindower:
New constraints that I just added to the implementation:
- plain machines may not be added to shared networks in addition to a standard private network (as we see no use case for this)
- machines of a shared network might have optionally a firewall

To enable internet access for machines in a shared network that have no firewall, we would need to be able to deploy a squid service that gets an IP of the shared network; so we'll have to adapt metal-ccm that it allows also private shared networks but this should not be allowed for everyone. 

/cc @Gerrit91

cmd/metal-api/internal/service/network-service.go
```diff
@@ -572,6 +577,9 @@ func (r networkResource) updateNetwork(request *restful.Request, response *restf
 	if requestPayload.Labels != nil {
 		newNetwork.Labels = requestPayload.Labels
 	}
+	if requestPayload.Shared != nil {
+		newNetwork.Shared = *requestPayload.Shared
+	}
```
mwindower, line 582: add validation: shared flag must not be set from shared to unshared!

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -1156,37 +1156,63 @@ func gatherNetworksFromSpec(ds *datastore.RethinkStore, allocationSpec *machineA
...
 		}
-		network, ok := specNetworks[ip.NetworkID]
-		if !ok {
-			return nil, fmt.Errorf("given ip %q is not in any of the given networks, which is required", ip.IPAddress)
+
+		if !pn.isPrimaryPrivate && !pn.network.Shared {
+			return nil, fmt.Errorf("private network is not allowed to be used because it is not marked as shared")
```
mwindower, line 1188: mention network that is not allowed

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -1156,37 +1156,63 @@ func gatherNetworksFromSpec(ds *datastore.RethinkStore, allocationSpec *machineA
...
-		s := ip.GetScope()
-		if s != metal.ScopeMachine && s != metal.ScopeProject {
-			return nil, fmt.Errorf("given ip %q is not available for direct attachment to machine because it is already in use", ip.IPAddress)
+
+		for _, ipString := range allocationSpec.IPs {
+			ip, err := ds.FindIPByID(ipString)
+			if err != nil {
+				return nil, err
+			}
+			if ip.ProjectID != allocationSpec.ProjectID {
```
mwindower, line 1196: if the network is shared and the allocation contains a firewall, projectID is allowed to differ

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -1272,14 +1298,14 @@ func makeMachineTags(m *metal.Machine, networks allocationNetworkMap, userTags [
 	labels := make(map[string]string)
 
 	for _, n := range networks {
-		if !n.isPrivate {
+		if !n.isPrimaryPrivate {
 			for k, v := range n.network.Labels {
 				labels[k] = v
 			}
 		}
```
mwindower, line 1305: check whether we can skip adding network labels to machines

cmd/metal-api/internal/service/machine-service_test.go
```diff
@@ -1253,7 +1253,104 @@ func Test_gatherNetworksFromSpec(t *testing.T) {
...
+					},
+				},
+			},
+			partition:              &testdata.Partition1,
+			partitionSuperNetworks: partitionSuperNetworks,
+			errRegex:               "machines are not allowed to be placed into multiple private networks",
+			wantErr:                true,
+		},
+		{
+			name: "try to add machine to multiple private networks which are not shared",
```
mwindower, line 1353: add test cases:
- add specific ip addresses for shared network machines
- use auto ip acquisition for shared network
- different firewall scenarios

cmd/metal-api/internal/service/machine-service.go
mwindower: This condition is still valid: the ip address should have a reference to the project where it is used!

With auto acquisition this is the case: `makeMachineNetwork` will insert the projectID of the allocation.
Using a specific IP is also okay: `metalctl network ip allocate --network shared --project consumer`

cmd/metal-api/internal/service/machine-service.go
mwindower: created an issue for that

cmd/metal-api/internal/service/machine-service.go
mwindower: This was actually a bug. `metalctl network ip allocate` for network with differing project id should only succeed for shared networks.

cmd/metal-api/internal/metal/machine.go
```diff
@@ -157,10 +157,108 @@ type MachineNetwork struct {
...
+	}
+	if nt == nil {
+		return nil, fmt.Errorf("could not determine network type out of flags, underlay: %v, privateprimary: %v, private: %v, shared: %v", mn.Underlay, mn.PrivatePrimary, mn.Private, mn.Shared)
+	}
+
+	if nt.Supported {
+		return nt, nil
+	}
+	// This is for machineNetworks from a Allocation which was before NetworkType was introduced.
+	// We guess based on unset fields not present at this time and therefor are set to false.
```
Gerrit91, line 253:
```suggestion
	// This is for machineNetworks from an allocation which was before NetworkType was introduced.
	// We guess based on unset fields not present at this time and therefore are set to false.
```

cmd/metal-api/internal/metal/machine.go
```diff
@@ -157,10 +157,108 @@ type MachineNetwork struct {
...
+	if mn.Private && !mn.PrivatePrimary && !mn.Shared && !mn.Underlay {
+		return &PrivatePrimaryUnshared, nil
+	}
```
Gerrit91, line 256: So, this is actually code that needs to be removed in a future release, right? Can we mention this somewhere?

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -65,28 +65,29 @@ type machineAllocationSpec struct {
...
 			break
 		}
 	}
-	if privateNetwork == nil {
-		return nil, fmt.Errorf("no private network contained")
+	if privatePrimaryNetwork == nil {
+		return nil, fmt.Errorf("no private primary network contained")
 	}
-	return privateNetwork, nil
+	return privatePrimaryNetwork, nil
```
Gerrit91, line 90: Mh... we now introduced all these nice network types but now we still use the old flags in our algorithms.

What about something like:

```go
func getPrivatePrimaryNetwork(networks allocationNetworkMap) (*allocationNetwork, error) {
	for _, n := range networks {
		if n.networkType == metal.PrivatePrimaryUnshared {
			return n, nil
		}
	}
	for _, n := range networks {
		if n.networkType == metal.PrivatePrimaryShared {
			return n, nil
		}
	}
	return nil, fmt.Errorf("no private primary network contained")
}
``` 
```

File: cmd/metal-api/internal/service/machine-service.go [hunk 12]
```diff
@@ -1146,24 +1150,25 @@ func gatherNetworksFromSpec(ds *datastore.RethinkStore, allocationSpec *machineA
...
-			ips:       []metal.IP{},
-			isPrivate: false,
+			network:     network,
+			auto:        auto,
+			ips:         []metal.IP{},
+			networkType: metal.External,
 		}
 
 		for _, privateSuperNetwork := range privateSuperNetworks {
 			if network.ParentNetworkID == privateSuperNetwork.ID {
```
Gerrit91, line 1160: Consider using guard function:

```go
if network.ParentNetworkID != privateSuperNetwork.ID {
    continue
}
```

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -1290,14 +1323,14 @@ func makeMachineTags(m *metal.Machine, networks allocationNetworkMap, userTags [
 	labels := make(map[string]string)
 
 	for _, n := range networks {
-		if !n.isPrivate {
+		if !n.networkType.PrivatePrimary && !n.networkType.Shared {
```
Gerrit91, line 1326: I hope we are gonna remove this in #120 right after this PR.

cmd/metal-api/internal/service/v1/machine.go
```diff
@@ -57,10 +57,17 @@ type MachineNetwork struct {
 	Vrf       uint     `json:"vrf" description:"the vrf of the allocated machine"`
 	// Attention, uint32 is converted to integer by swagger which is int32 which is to small to hold a asn
 	ASN                 int64    `json:"asn" description:"ASN number for this network in the bgp configuration"`
-	Private             bool     `json:"private" description:"indicates whether this network is the private network of this machine"`
 	Nat                 bool     `json:"nat" description:"if set to true, packets leaving this network get masqueraded behind interface ip"`
 	DestinationPrefixes []string `json:"destinationprefixes" modelDescription:"prefixes that are reachable within this network" description:"the destination prefixes of this network"`
-	Underlay            bool     `json:"underlay" description:"if set to true, this network can be used for underlay communication"`
+	NetworkType         string   `json:"networktype" description:"the network type"`
```
Gerrit91, line 62:
```suggestion
	NetworkType         string   `json:"networktype" description:"the network type, types can be looked up in the network package of metal-lib"`
```

Gerrit91 (approved):
Some last comments and improvement suggestions, but I am approving this already.

cmd/metal-api/internal/service/machine-service.go
Gerrit91: Mh... I guess, we can even remove this function after #120. So we can leave it as is.

cmd/metal-api/internal/service/machine-service.go
mwindower: good suggestion!

PR #105, opened 2020-09-21
Fix asn allocation

majst01:
Currently a ASN is calculated for a machine/firewall based on the last 2 octets of the ip address of the private network. This has several downsides:

- it will create asn collisions
- asn is dependent on distinct ip addresses per private network which makes it impossible to have identical private network ranges for different private networks

With this PR the ASN calculation is using the same mechanism as VRF allocation, it grabs a available integer from a precompiled integerpool which guarantees uniqueness.

To prevent collision with existing asn numbers of already allocated machines we  start allocation of asn with this method from a different range: 
4210000000
whereas the old scheme started at:
4200000000

### DIFF
--- a/cmd/metal-api/internal/datastore/integer.go
+++ b/cmd/metal-api/internal/datastore/integer.go
@@ -75,10 +75,6 @@ func (rs *RethinkStore) GetIntegerPool(pool IntegerPoolType) (*IntegerPool, erro
 // 	net add vxlan vxlan10 vxlan id
 //  <1-16777214>  :  An integer from 1 to 16777214
 //
-// FIXME Gerrit:
-// Aus dem Pool zieht man aber nicht die 4.000.000.000 raus.
-// Ist halt theoretisch möglich wie bei dem VRF 104009.
-// Man möchte 104010 einführen in seiner Infrastruktur und möglicherweise hat eine ASN dann diese ID herausgezogen und kein VRF. Weiß nicht, ob das so schlimm ist.^^
 // in order not to impact performance too much, we limited the range of integers to 2^17=131072,
 // which includes the range that we typically used for vrf names in the past.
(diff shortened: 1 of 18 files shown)


majst01:
Adoption is required once #108 is merged

cmd/metal-api/internal/datastore/integer.go
```diff
@@ -26,16 +26,24 @@ type Integerinfo struct {
...
+// - asn-id offset added to 4210000000 (ASNBase)
 //
 // the integer range has a vxlan-id constraint from Cumulus:
 // 	net add vxlan vxlan10 vxlan id
 //  <1-16777214>  :  An integer from 1 to 16777214
 //
+// FIXME Gerrit:
+// Aus dem Pool zieht man aber nicht die 4.000.000.000 raus.
+// Ist halt theoretisch möglich wie bei dem VRF 104009.
+// Man möchte 104010 einführen in seiner Infrastruktur und möglicherweise hat eine ASN dann diese ID herausgezogen und kein VRF. Weiß nicht, ob das so schlimm ist.^^
```
mwindower, line 43: This comment is partially superfluous with the seperation of pools, I guess?

To avoid conflicting VRF IDs with our well known ones (104009 and 104010), we could loop over the non private networks and acquire those IDs from the pool?

majst01: Will remove this comment regarding seperation.
avoiding that these "private" vrf ids are acquired is not possible as long as the integerpool does not support such protected areas.

majst01: after discussion with @mwindower there is no hard reason to reserve special vrf ids for private vrfs.

Gerrit91 (approved):

PR #106, opened 2020-09-22
first try

majst01:
Make the Integerpool configurable for different consumers. This is done by adding a name field to every ID in both tables integer and integerinfo.

This PR still lacks migration of existing data, unit tests failing, etc.


majst01:
obsolete through #108

PR #108, opened 2020-09-22
Seperate integer pool

majst01:
Make integerpool configurable in two aspekts:

- create a named pool, one for each purpose (vrf and asn in this case)
- min and max are configurable per pool

this is a required modification to prevent pool exhaustion by asn allocation in #105.

Downside: pool initialization is now twice as long, which make mini-lab initialization  ~20 sec slower.

### DIFF
--- a/cmd/metal-api/internal/datastore/integer.go
+++ b/cmd/metal-api/internal/datastore/integer.go
@@ -34,7 +34,7 @@ type integerinfo struct {
 
 // GetIntegerPool returns a named integerpool if already created
 func (rs *RethinkStore) GetIntegerPool(name string) (*IntegerPool, error) {
-	ip, ok := rs.IntegerPools[name]
+	ip, ok := rs.integerPools[name]
 	if !ok {
 		return nil, fmt.Errorf("no integerpool for %s created", name)
 	}
--- a/cmd/metal-api/internal/datastore/integer.go
+++ b/cmd/metal-api/internal/datastore/integer.go
@@ -8,6 +8,22 @@ import (
 	r "gopkg.in/rethinkdb/rethinkdb-go.v6"
 )
 
+// IntegerPoolType defines the name of the IntegerPool
+type IntegerPoolType string
+
+func (p IntegerPoolType) String() string {
+	return string(p)
+}
+
+const (
+	// VRFIntegerPool defines the name of the pool for VRFs
+	// this also defines the name of the tables
+	// FIXME, must be renamed to vrfpool later
+	VRFIntegerPool IntegerPoolType = "integerpool"
+	// ASNIntegerPool defines the name of the pool for ASNs
+	ASNIntegerPool IntegerPoolType = "asnpool"
+)
+
 var (
 	// VRFPoolRangeMin the minimum integer to get from the vrf pool
 	VRFPoolRangeMin = uint(1)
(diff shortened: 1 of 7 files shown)


cmd/metal-api/internal/datastore/integer.go
```diff
@@ -15,15 +15,32 @@ var (
 	IntegerPoolRangeMax = uint(131072)
 )
 
+// IntegerPool manages unique integers
+type IntegerPool struct {
+	tablename string
+	min       uint
```
Gerrit91, line 21: As soon as `min` and `max` are introduced as private fields, the variables `IntegerPoolRangeMin` and `IntegerPoolRangeMax` are kind of obsolete or at least the name is weak (because they are not a real maximum).

I think, it would be better to name them `VRFPoolMin/Max` and `ASNPoolMin/Max`.

cmd/metal-api/internal/datastore/integer.go
```diff
@@ -15,15 +15,32 @@ var (
...
 
-// Integerinfo contains information on the integer pool.
-type Integerinfo struct {
+// integerinfo contains information on the integer pool.
+type integerinfo struct {
 	IsInitialized bool `rethinkdb:"isInitialized" json:"isInitialized"`
 }
 
+// GetIntegerPool returns a named integerpool if already created
+func (rs *RethinkStore) GetIntegerPool(name string) (*IntegerPool, error) {
```
Gerrit91, line 36: Do we really need a pool map, which requires error handling on pool-user-side? Isn't it possible to just assign both pools to two variables and to access the directly?

cmd/metal-api/internal/datastore/integer.go
```diff
@@ -51,72 +68,79 @@ type Integerinfo struct {
...
-			return err
+			return nil, err
 		}
 	}
 
+	ip := &IntegerPool{
+		tablename: tablename,
+		min:       min,
+		max:       max,
+		rs:        rs,
```
Gerrit91, line 84: Do we need the `rs`? Isn't `tablename` enough for all actions after construction?

Gerrit91 (commented):
Looks quite good already. I left some comments open for discussion. :)

cmd/metal-api/internal/datastore/integer.go
majst01: Yes

cmd/metal-api/internal/datastore/integer.go
majst01: rs is required to actually read and write to the database:
https://github.com/metal-stack/metal-api/pull/108/files#diff-d354d7679d95da26fd3b929143a9a955R104

cmd/metal-api/internal/datastore/integer.go
majst01: i hate global variables

cmd/metal-api/internal/datastore/integer.go
majst01: done

cmd/metal-api/internal/datastore/integer.go
majst01: typed getter which creates the pool if not already initialized

cmd/metal-api/internal/datastore/integer.go
majst01: store term and session instead of rs in integerpool

mwindower:
lgtm

Gerrit91 (approved):

PR #111, opened 2020-10-01
enable update of fru when finalize allocation

kolsa:


Gerrit91:
We update this on `finalizeAllocation`? Why? :sweat_smile: 

I thought we want to allow the bmc-catcher to update these fields?

kolsa:
This is only an additional FRU update, that should not harm too much.
We should discuss if the bmc-catcher update is sufficient in all cases. If so, we can close this PR as well as the related ones in metal-core and metal-hammer

majst01:
I tend to see this also as a "Gürtel plus Hosenträger" approach. If we have bmc-catcher reporting these information, there should be no reason to double this update effort.

majst01 (commented):
I tend to close this in favour of the bmc-catcher approach

Gerrit91:
I'd also vote for closing because I think when the bmc-catcher starts to report these things like... every 10 minutes or so... it's enough. This makes the updates async to the machine allocation procedure, which I think is the way to go for keeping this kind of 
information up-to-date.

PR #113, opened 2020-10-05
Do not send error string.

Gerrit91:
We always return plaintext error responses, which makes handling errors client-side very hard and confusing (round-tripper does JSON Unmarshaling and throws away original error type from swagger client).

This is how a response currently looks like:

``` 
HTTP/2.0 404 Not Found
Content-Length: 77
Content-Type: text/plain; charset=utf-8
Date: Mon, 05 Oct 2020 12:05:00 GMT
Server: nginx/1.17.8
Strict-Transport-Security: max-age=15724800; includeSubDomains

{"statuscode":404,"message":"no network with id \"test-01\" found: NotFound"}
```

With this pull request it will look like this:

```
HTTP/2.0 404 Not Found
Content-Length: 84
Content-Type: application/json
Date: Mon, 05 Oct 2020 12:36:28 GMT
Server: nginx/1.17.8
Strict-Transport-Security: max-age=15724800; includeSubDomains

{
 "statuscode": 404,
 "message": "no network with id \"test-02\" found: NotFound"
}
``` 

Then, we can start handle errors in clients properly like this (requires to remove the RoundTripper from metal-go though, because it destroys the original error type):

```go
resp, err := driver.NetworkGet(*nar.ID)
if err != nil {
	switch e := err.(type) {
	case *networkmodel.FindNetworkDefault:
                // this part in the code can now be reached, which is impossible with the current impl
		if e.Code() != http.StatusNotFound {
			return fmt.Errorf("network get error:%v", e.Error())
		}
	default:
		return fmt.Errorf("unexpected error on network get:%v", err)
	}
}
```

(should also remove confusing "TextUmarshaller missing" issues!)

### DIFF
--- a/cmd/metal-api/internal/service/service.go
+++ b/cmd/metal-api/internal/service/service.go
@@ -1,7 +1,6 @@
 package service
 
 import (
-	"encoding/json"
 	"fmt"
 	"net/http"
 	"strings"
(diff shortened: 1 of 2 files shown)


majst01 (approved):
Nice Catch

kolsa (approved):

PR #119, opened 2020-10-13
Fix machine allocation, liveliness is not in machine entity.

Gerrit91:


cmd/metal-api/internal/datastore/machine.go
```diff
@@ -442,19 +436,40 @@ func (rs *RethinkStore) FindWaitingMachine(partitionid, sizeid string) (*metal.M
...
+		if !ok {
+			rs.SugaredLogger.Errorw("cannot find machine provisioning event container", "machine", m, "error", err)
+			// fall through, so the rest of the machines is getting evaluated
+			continue
+		}
+		switch ec.Liveliness {
+		case metal.MachineLivelinessAlive:
+			available = append(available, m)
+		}
+	}
```
mwindower, line 466: check whether we can get the same behavior with a table join?

majst01: We add a hint that this might be a optimization

majst01 (approved):

PR #122, opened 2020-10-15
fix waitServer test by using in-memory connections. Closes #118

### LINKED ISSUE #118: Tests failing in CI from time to time
We already had some issues with this in the past. Those are tests where the gRPC wait server is involved.

Here is an example of how it looks like:

https://github.com/metal-stack/metal-api/pull/115/checks?check_run_id=1242007616

kolsa:


majst01:
LGTM can we merge this ?

Gerrit91 (approved):

Gerrit91:
I think yes!

cmd/metal-api/internal/grpc/waitServer.go
```diff
@@ -247,7 +247,9 @@ func (s *WaitServer) Wait(req *v1.WaitRequest, srv v1.Wait_WaitServer) error {
 	defer func() {
 		s.queueLock.Lock()
 		delete(s.queue, machineID)
-		close(can)
+		if err != nil {
```
majst01, line 250: cant immediately see where this error is coming from

kolsa: This happened a few times during 100+ subsequent test runs

kolsa: I'll revert this for now and try to test such a scenario

majst01: OK, now the tests fail, this is weird. so we need the errcheck

Gerrit91:
Probably not required anymore due to #131.

PR #124, opened 2020-10-20
Introduce network type enum

mwindower:


cmd/metal-api/internal/metal/machine.go
```diff
@@ -165,6 +165,41 @@ type MachineNetwork struct {
...
+const (
+	PrivatePrimaryUnshared NetworkType = "privateprimaryunshared"
+	PrivatePrimaryShared   NetworkType = "privateprimaryshared"
+	PrivateSecondaryShared NetworkType = "privatesecondaryshared"
+	Underlay               NetworkType = "underlay"
+)
+
+// NetworkType determines the network type based on the flags stored in the db entity.
+func (mn *MachineNetwork) NetworkType() (*NetworkType, error) {
+	if mn.Underlay && !mn.PrivatePrimary && !mn.Private && !mn.Shared {
```
majst01, line 180: if underlay is set to true i would set networktyp to underlay regardles of other flags ?

mwindower: if other flags contradict the underlay flag I would propose to raise an error.

majst01:
Love the idea of still preserving the boolean fields to have a kind of backward compatibility.

majst01 (approved):

PR #126, opened 2020-10-23
fix network release when freeing machine asynchronously

kolsa:


Gerrit91:
I talked to @ulrichSchreiner about this PR and we both think that this is not fixing anything. 

I remember I also thought about the same thing when I reviewed the async-actor back then. But when the machine is given to the machine releaser, this object gets serialized and stored into NSQ and then all subsequent actions of the async actor will be called with the deserialized, unmodified machine object. Therefore, there will be no race conditions.

If you still think that this is indeed a bug, I need a proof. :)

Gerrit91:
Please re-open if you still think this is an issue.

kolsa:
Make sense and should be fine.
But then we should fix at least the race condition in `TestFreeMachine` without this change.

PR #134, opened 2020-11-06
configurable reserve of machines per size per partition

majst01:
Currently consumers, like cluster-api, of metal-api actually just check if enough workers for the current request are available. With this approach we will end up with a situation that all workers are allocation in a partition.
But this will prevent further actions like rolling updates of worker nodes or firewalls.
To prevent this a configurable amount of machines must stay as reserved to be able to perform these updates.
The reserved machines are now configurable per size and partition.
But this implementation only performs the configuration of these reservation and does not enforce them. This enforcement is still a responsibility of the caller side.


majst01:
we should add a fixed amount on top of this value during cluster creation.

Gerrit91:
Somehow I don't understand what this is doing. Any explanation such that I can understand it?

Gerrit91:
We've decided to implement this directly in the cloud-api (probably in config file) because this approach has no direct advantage for adopters.

PR #137, opened 2020-11-12
Swagger field parentnetworkid in network response missing optional:true.

Gerrit91:
This stuff is only problematic for metal-python, it is complaining about `None` being returned by the API even though it should be a string.


majst01:
Will change metal-go, needs tests in consumers.

Gerrit91:
I can test this. 

However, the impact should be pretty much identical to what we already saw in https://github.com/metal-stack/metal-api/pull/81. For metal-go it does not seem to matter when the spec is slightly wrong in that regard (spec says we are returning a string but in our struct we use a pointer and can return `None` occasionally). The metal-python client does more validations (probably because of missing type-safety) and then fails.

Gerrit91:
Tested with metalctl from https://github.com/metal-stack/metalctl/pull/64 (based on https://github.com/metal-stack/metal-go/pull/50). Following things work as before:

- network list
- network search (with and without parent id)
- network create
- network rm

majst01 (approved):

PR #147, opened 2020-11-25
Fix issue where IP can be allocated for arbitrary projects.

Gerrit91:
Fixes an issue that allowed allocating IP addresses in network with a different project than the IP allocation.


Gerrit91:
Was a mistake, this works fine, the misleading behavior was for MPLS networks.

PR #148, opened 2020-11-27
Do not hide images on machine create which are already expired

majst01:


Gerrit91 (approved):
Reasoning behind this also was that the API should not filter out expired images because they actually exist in the image list, but then suddenly they disappear on more specific calls, which is inconsistent and is also very likely to break customer experience.

PR #149, opened 2020-11-27
Update to go 1.15 in module, use alpine 3.12, remove redoc serve path

majst01:
@droid42 our image from where we got the redoc-cli was not modified/updated since years, created a redoc-cli installation stage directly. How can we access the redoc gui  and see what changed ?


majst01:
Should we move the redoc generated swagger documentation move to the docs project ?
Votes ?

Gerrit91:
Mh, personally, I don't really think the redoc endpoint is necessary and I have never met someone who used it. To me it's just an increase of the API's attack surface. I do provide the redoc documentation already for every metal-stack release.

Gerrit91 (approved):

majst01:
But where is the redoc documentation ?

Gerrit91:
https://docs.metal-stack.io/stable/apidocs/apidocs/

majst01:
OK, perfect, so i will remove the apidocs from the metal-api container.

PR #152, opened 2021-01-01
use go-ipam with ipv6 support

majst01:
This enables basic ipv6 support from go-ipam
https://github.com/metal-stack/go-ipam/pull/36

https://github.com/gardener/gardener/issues/4453

- [ ] run test in mini-lab, ensure all works as before with ipv4

TODO:
- [x] availableips and availablesmallestprefixes are always 0 only on ipv6 networks
- [ ] should we allow networks with mixed address families ? E.g. the internet is made out of both. This way we must add a --ipv6 flag to `network ip allocate` to get a ipv6 address from a mixed af network.


Gerrit91 (approved):
Looks good, let's try it in the mini-lab.

cmd/metal-api/internal/service/network-service.go
```diff
@@ -301,18 +322,37 @@ func (r networkResource) createNetwork(request *restful.Request, response *restf
...
 			if err != nil {
 				if !metal.IsNotFound(err) {
 					if checkError(request, response, utils.CurrentFuncName(), err) {
 						return
 					}
 				}
 			} else {
-				if checkError(request, response, utils.CurrentFuncName(), fmt.Errorf("partition with id %q already has a private super network", partition.ID)) {
-					return
+				existingsuper := nw.Prefixes[0].String()
```
majst01, line 338: check before len(nw.Prefixes)

Gerrit91:
Included in #549.

PR #154, opened 2021-01-12
implement machine deletion

### LINKED ISSUE #153: Possibility to delete orphaned/repaired machines from the api/database
Sometimes it is necessary to remove a repaired machine from the api/database. The UUID will change when a mainboard or a machine is replaced by the vendor. An admin needs the possibility to remove this orphaned machine from the API because the old UUID will be never registered again, something like:

`metalctl machine remove/repair bbbe3200-6a76-11e9-8000-ac1f6b7bece4`

For safety reasons, this should be not possible with a "phoned home" machine without issues.

muhittink:
closes #153

### DIFF
--- a/cmd/metal-api/internal/service/machine-service.go
+++ b/cmd/metal-api/internal/service/machine-service.go
@@ -216,6 +216,16 @@ func (r machineResource) webService() *restful.WebService {
 		Returns(http.StatusOK, "OK", v1.MachineResponse{}).
 		DefaultReturns("Error", httperrors.HTTPErrorResponse{}))
 
+	ws.Route(ws.DELETE("/{id}").
+		To(admin(r.deleteMachine)).
+		Operation("deleteMachine").
+		Doc("deletes a machine from the database").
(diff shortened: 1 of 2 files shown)


cmd/metal-api/internal/service/machine-service.go
```diff
@@ -1486,6 +1496,66 @@ func (r machineResource) freeMachine(request *restful.Request, response *restful
...
+	}
+	logger := utils.Logger(request).Sugar()
+
+	if m.Allocation != nil {
+		checkError(request, response, utils.CurrentFuncName(), fmt.Errorf("cannot delete machine that is allocated"))
+		return
+	}
+
+	ec, err := r.ds.FindProvisioningEventContainer(id)
+
```
majst01, line 1513: remove the whitespace between function call and the error handling

majst01 (approved):

majst01:
Please raise required PRs for metal-go and metalctl.

PR #156, opened 2021-01-25
Add data store connect opts.

Gerrit91:
The intention is to run database initialization or the database admin user only when it's required.

Running database initialization always can be problematic when scenarios like the following occur:

- The rethinkdb does not come during the deployment for some reason
- Commands from cron jobs like evaluateMachineLiveliness, resurrectDeadMachines, ... queue up in Kubernetes
- When the database comes up all the pods start execution at the same time
- If there was a change in the tables, it can happen that tables are created multiple times -> Rethinkdb is broken

Generally, we should configure the deployment such that only the init-db pre-deployment-hook initializes the database because the hook is guaranteed to only run one at a time.

Also, all tasks should do user demotion, but for the db-init command and the migrate command this is not required because they do not affect possible migrations while running.

### DIFF
--- a/cmd/metal-api/internal/datastore/integer.go
+++ b/cmd/metal-api/internal/datastore/integer.go
@@ -115,7 +115,7 @@ func (ip *IntegerPool) String() string {
 func (ip *IntegerPool) initIntegerPool(log *zap.SugaredLogger) error {
 	var result integerinfo
 	err := ip.infoTable.ReadOne(&result, ip.session)
-	if err != nil {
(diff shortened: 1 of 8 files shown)


Gerrit91:
Had to refactor the integer pool a bit as otherwise calling the init functions would be required for every command that uses the database. It's better now I guess because you also do not need `RenewSession` on the integer pools anymore.

Gerrit91:
Verified fresh installation and upgrade through mini-lab.

majst01 (approved):
dont forget to merge helm chart adoptions

PR #159, opened 2021-01-28
Fix labels field in createNetwork not processed

mwennrich:


cmd/metal-api/internal/service/network-service.go
```diff
@@ -227,6 +227,14 @@ func (r networkResource) createNetwork(request *restful.Request, response *restf
...
 		vrfShared = *requestPayload.VrfShared
 	}
+	labels := make(map[string]string)
+	if requestPayload.Labels != nil {
+		labels = requestPayload.Labels
+	}
+	shared := false
+	if requestPayload.Shared != nil {
+		shared = *requestPayload.Shared
+	}
```
Gerrit91, line 237: I would assume, the shared flag should not be possible to be set by the user at this place and hence not be part of the create request. The creation endpoint is only for managing super networks, external networks, ... which should never have the shared property set to true (can be looked up when they are mapped to machine networks [here](https://github.com/metal-stack/metal-api/blob/v0.12.1/cmd/metal-api/internal/metal/machine.go#L180-L229)).

For the update request we should have a validation that checks that only private networks can switch from unshared to shared.

@mwindower Are my assumptions correct?

@mwennrich If you want this PR to go through fast, you can just remove the `shared` property and only do the labels. We can also follow-up this discussion in a separate PR / issue.

mwennrich: removed Shared field

Gerrit91: https://github.com/metal-stack/metal-api/issues/160

Gerrit91 (approved):

PR #162, opened 2021-02-02
add upload/update bios/bmc machine endpoints

kolsa:


majst01:
I would not expose a upload bmc/bios update functionality, i would only allow the configuration of a downloadable base URL of the bmc/bios version. If under this url nothing resides, update skipes.

I would also just rely on plain https for the download, no s3 client whatsoever, this will reduce complexity a lot in all related components (e.g. go-hal, metal-core etc.)

Gerrit91:
The feature is great!

Though, from the deployment perspective it would also be easier not to rely on S3 compatible backends as a user. 

I think it would enhance usability if this entire thing would feel a bit like managing our OS images such that we have a common UX all over. Not sure if we need to provide authentication mechanisms for the URLs because artifacts need to be hidden from public access? Maybe a small architectural draft with requirements and implementation guidelines for this entire thing would be helpful. If you want we can schedule a meeting for that, too.

kolsa:
The upload feature encapsulates the required directory and file naming conventions and even creates missing buckets/directories, which otherwise might be cumbersome in future. Since both endpoints are only accessible by editors, why not having also this upload feature?
I agree that the updated files do not have to be hidden from public.

cmd/metal-api/main.go
```diff
@@ -173,6 +174,12 @@ func init() {
 
 	rootCmd.Flags().StringP("base-path", "", "/", "the base path of the api server")
 
+	rootCmd.Flags().StringP("s3-region", "", "", "the region of the s3 server")
+	rootCmd.Flags().StringP("s3-address", "", "", "the address of the s3 server")
+	rootCmd.Flags().StringP("s3-key", "", "", "the key of the s3 server")
+	rootCmd.Flags().StringP("s3-secret", "", "", "the secret of the s3 server")
+	rootCmd.Flags().BoolP("s3-hostname-immutable", "", false, "whether the hostname is immutable or not")
```
majst01, line 181: remove this

kolsa:
As agreed during our meeting here is a sketch of this functionality:

To not offend against SSOT and also reduce complexity for a feature that is not used very often, I decided to model the BIOS/BMC update functionality through the filesystem only (no DB table overhead).

Thus we need a directory structure that mirrors the distinct board types of the hardware vendors, which contains the BIOS/BMC update files. It must also be possible to query all available BIOS/BMC versions of a certain board.

The current solution looks like this:
```
<bucket named after vendor>
|---- bios|bmc
       |---- <board>
              |---- <update files named after version>
```

Let's assume that we have three BIOS updates (version `3.1` and `3.3`) applicable to all Supermicro Machine with board type `X11DPT-B`, the final directory structure looks like this:
```
supermicro/bios/X11DPT-B
|--- 3.1
|--- 3.3
```

Query example for one of these machines with UUID `00000000-0000-0000-000000000123`:
```
$ metalctl machine update bios 00000000-0000-0000-000000000123
3.1 (current)
3.3
```

Update example:
```
$ metalctl machine update bios --revision 3.3 --yes-i-really-mean-it 00000000-0000-0000-000000000123
```

Upload example of a new BIOS update.
It is required to manually download this BIOS update and extract the corresponding file, e.g. `X11DPTB0.B03` (version `3.4`):
```
$ metalctl machine upload bios --vendor supermicro --board x11dpt-b --revision 3.4 --yes-i-really-mean-it X11DPTB0.B03
```

It would be possible to also implement this by a machine prototype:
```
$ metalctl machine upload bios --revision 3.4 --file X11DPTB0.B03 --yes-i-really-mean-it 00000000-0000-0000-000000000123
```

kolsa:
We also agreed to hide the update files from public. The current implementation takes care of this.

cmd/metal-api/main.go
kolsa: s3-hostname-immutable removed

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -399,26 +399,26 @@ func (r machineResource) webService() *restful.WebService {
 		Returns(http.StatusOK, "OK", v1.MachineResponse{}).
 		DefaultReturns("Error", httperrors.HTTPErrorResponse{}))
 
-	ws.Route(ws.POST("/upload/bios").
+	ws.Route(ws.POST("/upload/bios/{vendor}/{board}/{revision}").
```
majst01, line 402: I would rather combine these endpoint with bmc and create one firmware-service where one upload endpoint exists which contains bmc/bios/networkcard firmware blobs.

WDYT ?

kolsa: Sounds good, I'll do so after my tests

majst01:
Great !

cmd/metal-api/internal/service/v1/s3.go
```diff
@@ -0,0 +1,53 @@
...
+			return aws.Endpoint{
+				PartitionID:       "aws",
+				URL:               c.Url,
+				SigningRegion:     region,
+				HostnameImmutable: true,
+			}, nil
+		}
+		// returning EndpointNotFoundError will allow the service to fallback to it's default resolution
+		return aws.Endpoint{}, &aws.EndpointNotFoundError{}
+	})
```
Gerrit91, line 44: I guess you tried this against a GCP Bucket? For this, I think we have simpler connection params that we use for the image-cache-sync: https://github.com/metal-stack/metal-image-cache-sync/blob/master/cmd/main.go#L175-L191

cmd/metal-api/internal/service/v1/s3.go
```diff
@@ -0,0 +1,53 @@
+package v1
```
Gerrit91, line 1: Currently we only store request and response definitions in here. I think you can just move it to a separate s3 package.

At some point in time we should probably move the v1 package away from `internal` to `pkg/api` or something... but that can be another PR in the future.

Gerrit91:
Please also merge latest master into this branch and resolve merge conflicts.

cmd/metal-api/internal/service/integration_test.go
```diff
@@ -1,3 +1,4 @@
+//nolint:unused
```
Gerrit91, line 1: I think this arises due to the `// +build integration` tag? It's also possible to run goloangci-lint with `--build-tags`, which potentially resolves this problem.

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -2179,3 +2258,102 @@ func (s machineAllocationSpec) noautoNetworkN() int {
...
+	_, err = io.Copy(buf, file)
+	if checkError(request, response, utils.CurrentFuncName(), file.Close()) ||
+		checkError(request, response, utils.CurrentFuncName(), err) {
+		return
+	}
+
+	resp, err := r.s3Client.ListBuckets(context.Background(), nil)
+	if checkError(request, response, utils.CurrentFuncName(), err) {
+		return
+	}
```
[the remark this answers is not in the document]
Gerrit91: It has the potential to be a little easier and more efficient with a construct like this:

```
func EnsureBackupBucket(bucket string) error {
	cparams := &s3.CreateBucketInput{
		Bucket: bucket,
	}

	_, err := r.s3Client.CreateBucket(cparams)
	if err != nil {
		if aerr, ok := err.(awserr.Error); ok {
			switch aerr.Code() {
			case s3.ErrCodeBucketAlreadyExists:
			case s3.ErrCodeBucketAlreadyOwnedByYou:
			default:
				return err
			}
		} else {
			return err
		}
	}

       return nil
}
```

And then there seems to be a new uploader, which can intelligently upload files in multiple go routines in shorter time: https://docs.aws.amazon.com/sdk-for-go/api/service/s3/#hdr-Upload_Managers

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -1957,6 +2027,15 @@ func (r machineResource) chassisIdentifyLEDOff(request *restful.Request, respons
 	r.machineCmd("chassisIdentifyLEDOff", metal.ChassisIdentifyLEDOffCmd, request, response, request.QueryParameter("description"))
 }
 
+func (r machineResource) updateFirmware(request *restful.Request, response *restful.Response) {
+	var p v1.MachineUpdateFirmware
+	err := request.ReadEntity(&p)
+	if checkError(request, response, utils.CurrentFuncName(), err) {
+		return
+	}
+	r.machineCmd("updateFirmware", metal.UpdateFirmwareCmd, request, response, p.Kind, p.Revision, p.Description, r.s3Client.Url, r.s3Client.Key, r.s3Client.Secret)
```
Gerrit91, line 2036: Would it make sense to check if the firmware exists before handing it down to the metal-core?

Gerrit91 (commented):
Thanks for processing the comments! :+1:  

Just added some final thoughts, I guess. But it's already in a state where we can try it in our test environment? I can help with it if you want, then we can finish up the deployment part quickly.

cmd/metal-api/main.go
```diff
@@ -230,10 +236,10 @@ func init() {
 	rootCmd.Flags().StringP("hmac-view-lifetime", "", "30s", "the timestamp in the header for the HMAC must not be older than this value. a value of 0 means no limit")
 
 	rootCmd.Flags().StringP("hmac-edit-key", "", "must-be-changed", "the preshared key for hmac security for a editing user")
-	rootCmd.Flags().StringP("hmac-edit-lifetime", "", "30s", "the timestamp in the header for the HMAC must not be older than this value. a value of 0 means no limit")
+	rootCmd.Flags().StringP("hmac-edit-lifetime", "", "90s", "the timestamp in the header for the HMAC must not be older than this value. a value of 0 means no limit")
 
 	rootCmd.Flags().StringP("hmac-admin-key", "", "must-be-changed", "the preshared key for hmac security for a admin user")
-	rootCmd.Flags().StringP("hmac-admin-lifetime", "", "30s", "the timestamp in the header for the HMAC must not be older than this value. a value of 0 means no limit")
+	rootCmd.Flags().StringP("hmac-admin-lifetime", "", "90s", "the timestamp in the header for the HMAC must not be older than this value. a value of 0 means no limit")
```
Gerrit91, line 242: It's a little unfortunate that the firmware upload then depends on the speed of the internet connection in order not to fail the upload request... would be good if there was a chance to exclude the upload from the HMAC lifetime.

Made a follow-up issue here: https://github.com/metal-stack/metal-api/issues/177

Gerrit91 (approved):

PR #166, opened 2021-02-05
Swallow phoned home events while machine is shutting down.

Gerrit91:
With https://github.com/metal-stack/metal-api/pull/141 we introduced emission of a planned reboot event on machine deletion, which has couple of positive effects. Unfortunately, during the shutdown of a machine (which takes some time), the machine sets the status back to "phoned home", which makes the machine appear in the list of "faulty" machines.


Gerrit91:
We also have an odd state when machine reset does not work for some reason and machine is just hanging there, probably still serving client workload but having no user allocation. But this state can occur with or without this PR. Not sure if there's a good way to detect this state.

Gerrit91:
Superseded by https://github.com/metal-stack/metal-api/pull/265.

PR #167, opened 2021-02-08
Detect failed machine reclaim

Gerrit91:
Detects a machine state described in #166 where machine reclaim did not work (allocation is gone, still phoning home, but restart does not happen). 

This state can easily be created in the mini-lab because where we do not have a BMC. A user has to manually restart the machine to boot it back into the metal-hammer. Effectively though, a user could also still use the machine, which is something we should prevent. So, this state helps operations to identify such machines.

Looks like this in the mini-lab:

``` 
m machine ls
ID                                                      LAST EVENT      WHEN    AGE     HOSTNAME        PROJECT SIZE            IMAGE   PARTITION 
e0ab02d2-27cd-5a5e-8efc-080ba80cf258                    Waiting         35s                                     v1-small-x86            vagrant  
2294c949-88f6-5390-8154-fa53d93a3313            ❓      Planned Reboot  9s                                      v1-small-x86            vagrant  
```

### DIFF
--- a/cmd/metal-api/internal/service/machine-service.go
+++ b/cmd/metal-api/internal/service/machine-service.go
@@ -1787,7 +1787,7 @@ func (r machineResource) provisioningEventForMachine(machineID string, e v1.Mach
 		ec.Liveliness = metal.MachineLivelinessAlive
 	} else {
 		if event.Event == metal.ProvisioningEventPhonedHome && len(ec.Events) > 0 && ec.Events[0].Event == metal.ProvisioningEventPlannedReboot {
-			// machine free emits a planned reboot event, prevent the machine to set state back to phoned home while shutting down
+			// machine free emits a planned reboot event, now prevent the machine to set state back to phoned home while shutting down
 			zapup.MustRootLogger().Sugar().Debugw("swallowing phone home event after machine deletion", "id", ec.ID)
 		} else {
 			ec.Events = append([]metal.ProvisioningEvent{event}, ec.Events...)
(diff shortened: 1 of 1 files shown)


Gerrit91:
Different issue, but while playing around with this I also saw this happening:

``` 
{
"event": "Phoned Home" ,
"message": "provisioned since 2021-02-08T08:11:45Z" ,
"time": Mon Feb 08 2021 08:32:18 GMT+00:00
} ,
{
"event": "Waiting" ,
"message": "waiting for allocation" ,
"time": Mon Feb 08 2021 08:32:14 GMT+00:00
} ,
{
"event": "Registering" ,
"message": "start registering" ,
"time": Mon Feb 08 2021 08:32:14 GMT+00:00
} ,
{
"event": "Preparing" ,
"message": "starting metal-hammer version:"v0.7.5 (4e6b4a84), tags/v0.7.5-0-g4e6b4a8, 2020-11-23T11:47:38Z"" ,
"time": Mon Feb 08 2021 08:32:03 GMT+00:00
} ,
{
"event": "PXE Booting" ,
"message": "machine sent extended dhcp request" ,
"time": Mon Feb 08 2021 08:31:54 GMT+00:00
} ,
...
``` 

A "Phoned Home" event arrives while being in the wait state.

After "force off" on the VM and restarting (which of course is much faster than on real hardware, so it's probably more likely to happen) it seems like metal-core still sends a phoned home message even after the machine went into waiting state. Wonder how this can happen, but probably it's a metal-core thing, where events are getting buffered?

majst01:
I would like to postpone the solution of these edge-cases and prevent that the current event handling logic gets even more complex.
For me the whole event handling logic should be expressed with a state machine. 
There are already quite simple but powerful state machine libraries for go available like: https://github.com/looplab/fsm

Gerrit91:
I agree with you, would be much better to have an adequate library that makes this complex cases better to test. But I am pretty sure that the rewrite will not happen anytime soon and it's a bit sad that we do not clearly show operators when there was an issue reclaiming a machine from a user because the user gets compute resources for free in this case.

majst01:
Bonus of using a state machine like the one above, you can create a graphical visualization of the state diagram configured. Example: 
StateMachine

majst01:
Another quite new implementation of a go idiomatic state machine and zero dependencies:
https://github.com/qmuntal/stateless

majst01:
Implementing a State Machine by our self would also not be to hard and may end up in ~200 LOC

https://venilnoronha.io/a-simple-state-machine-framework-in-go

majst01:
> Implementing a State Machine by our self would also not be to hard and may end up in ~200 LOC
> 
> https://venilnoronha.io/a-simple-state-machine-framework-in-go

Stored this code here: https://github.com/majst01/ssm

majst01:
There is also a very new and simple impl: https://github.com/cocoonspace/fsm

Gerrit91:
Superseded by #265.

PR #170, opened 2021-02-23
Improve performance by using errors.New rather than fmt.Errorf where …

kolsa:
…no formatting is needed; avoid variable names hiding keywords and package names

### DIFF
--- a/cmd/metal-api/internal/metal/errors.go
+++ b/cmd/metal-api/internal/metal/errors.go
@@ -1,16 +1,14 @@
 package metal
 
 import (
-	"fmt"
-
 	"github.com/pkg/errors"
 )
 
 var (
-	errNotFound = fmt.Errorf("NotFound")
-	errConflict = fmt.Errorf("Conflict")
+	errNotFound = errors.New("NotFound")
+	errConflict = errors.New("Conflict")
 	// TODO refactor implentations of fmt.Errorf to metal.Internal() in datastore and service
-	errInternal = fmt.Errorf("Internal")
+	errInternal = errors.New("Internal")
 )
 
 // NotFound creates a new notfound error with a given error message.
--- a/cmd/metal-api/internal/service/machine-service.go
+++ b/cmd/metal-api/internal/service/machine-service.go
@@ -441,7 +441,7 @@ func (r machineResource) setMachineState(request *restful.Request, response *res
 
 	if machineState != metal.AvailableState && requestPayload.Description == "" {
 		// we want a cause why this machine is not available
-		if checkError(request, response, utils.CurrentFuncName(), fmt.Errorf("you must supply a description")) {
+		if checkError(request, response, utils.CurrentFuncName(), errors.New("you must supply a description")) {
 			return
 		}
 	}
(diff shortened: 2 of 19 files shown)


kolsa:
To be continued...

majst01:
Thanks for this work, we should address in this all occurrences of `"%v", err` which should be `"%w", err` according to `golangci-lin run -p bugs`, see: https://github.com/metal-stack/metal-api/issues/165

kolsa:
There is one final issue `G107: Potential HTTP request made with variable url (gosec)`, which potentially enables authorized users to [SSRF](https://www.acunetix.com/blog/articles/server-side-request-forgery-vulnerability/) attack metal-api.
Let's discuss how to deal with that.

cmd/metal-api/internal/metal/errors.go
```diff
@@ -20,7 +18,7 @@ func NotFound(format string, args ...interface{}) error {
 
 // IsNotFound checks if an error is a notfound error.
 func IsNotFound(e error) bool {
-	return errors.Cause(e) == errNotFound
+	return errors.Is(errors.Cause(e), errNotFound)
```
majst01, line 21: are these functions covered by a  test ? if yes this is great

majst01 (commented):
please check if the error matching is covered by a unit test, other than that LGTM

cmd/metal-api/internal/metal/errors.go
kolsa: Done.

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -1826,7 +1830,7 @@ func MachineLiveliness(ds *datastore.RethinkStore, logger *zap.SugaredLogger) er
 		case metal.MachineLivelinessDead:
 			dead++
 			p.Dead++
-		default:
+		case metal.MachineLivelinessUnknown:
```
majst01, line 1833: Is this the only case we count as unknown ?

cmd/metal-api/main.go
```diff
@@ -493,14 +493,15 @@ func initIpam() {
 		if err != nil {
 			logger.Errorw("cannot connect to db in root command metal-api/internal/main.initIpam()", "error", err)
 			time.Sleep(3 * time.Second)
-			goto tryAgain
```
majst01, line 496: now metal-api will not start anymore if ipamdb is not there ?

majst01: is probably ok

go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/Microsoft/hcsshim v0.8.10 // indirect
 	github.com/containerd/continuity v0.0.0-20201119173150-04c754faca46 // indirect
 	github.com/dustin/go-humanize v1.0.0
+	github.com/emicklei/go-restful v2.15.0+incompatible
```
majst01, line 10: There must be a import with the old go-restful somewhere lingering around, this must be converted to use v3 as well

majst01: or go mod tidy is missing

cmd/metal-api/main.go
kolsa: Yeah, but there's no functional change here at all

go.mod
kolsa: You're right, go mod tidy did the trick

cmd/metal-api/internal/service/machine-service.go
kolsa: Yep

majst01 (approved):

Gerrit91 (approved):

Gerrit91:
Mh... on the master branch this failed. We need a follow-up.

https://github.com/metal-stack/metal-api/actions/runs/638652346

PR #174, opened 2021-03-11
169 double ips

kolsa:
This MR avoids potential double ips coming from allocation spec. However, this does not solve the issue.
I couldn't spot the root cause yet, since auto IP allocation only ever asks ipamer for a single ip per network. I verified that all such networks are distinct.
To be continued...


majst01:
closing in favour of #189

PR #176, opened 2021-03-14
Support for multiple issuers / generic token issuer

droid42:
Implements #175 Support for multiple issuers / generic token issuer

Adds tenant endpoint to get tenant list via metal-go and integrates masterdata-api backed multiissuercache in a backwards compatible way.


cmd/metal-api/main.go
```diff
@@ -242,6 +246,7 @@ func init() {
 	rootCmd.Flags().StringP("hmac-admin-lifetime", "", "90s", "the timestamp in the header for the HMAC must not be older than this value. a value of 0 means no limit")
 
 	rootCmd.Flags().StringP("provider-tenant", "", "", "the tenant of the maas-provider who operates the whole thing")
+	rootCmd.Flags().StringP("issuerCacheInterval", "", "30m", "issuer cache invalidation interval, e.g. 60s, 30m, 2h45m - default 30m")
```
Gerrit91, line 249: The other parameters are hyphenated, this one is camel-cased. Maybe should also align with the others to make it look more consistent from the outside.

droid42: Good point, done

Gerrit91 (approved):

PR #182, opened 2021-03-26
Dynamically calculate rack ID in machine registration

### LINKED ISSUE #181: Machine cannot register
Sometimes it happens that a machine cannot register with metal-hammer with the following error:
```
{"level":"error","time":"2021-03-23T12:38:42.330Z","caller":"service/service.go:90","message":"service error","rqid":"635d9848b8f62ae85fc3917eac179a71","remoteaddr":"10.20.4.6","method":"POST","uri":"/metal/v1/machine/register","route":"/metal/v1/machine/register","useremail":"metal-edit@metal-stack.io","operation":"registerMachine","status":422,"error":"machine 00000000-0000-0000-0000-3cecef1a9daa is not connected to exactly two switches, found connections to 0 switches","service-caller":"machine-service.go:626","resp":"machine 00000000-0000-0000-0000-3cecef1a9daa is not connected to exactly two switches, found connections to 0 switches (422)"}
```

But later it will work

mwindower:
Fixes a bug leading to machine registration failure when using multiple leaf pairs in the same partition.

Fixes #181

### DIFF
--- a/cmd/metal-api/internal/service/machine-service.go
+++ b/cmd/metal-api/internal/service/machine-service.go
@@ -592,10 +592,7 @@ func (r machineResource) registerMachine(request *restful.Request, response *res
 			Allocation:  nil,
 			SizeID:      size.ID,
 			PartitionID: partition.ID,
-			// the RackID is calculated by the switch connections
-			// a metal-core instance might respond to pxe requests from all racks
-			//RackID:      requestPayload.RackID,
-			Hardware: machineHardware,
+			Hardware:    machineHardware,
 			BIOS: metal.BIOS{
 				Version: requestPayload.BIOS.Version,
 				Vendor:  requestPayload.BIOS.Vendor,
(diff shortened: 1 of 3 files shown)


majst01 (approved):
Before merging, test in fra required

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -591,8 +592,10 @@ func (r machineResource) registerMachine(request *restful.Request, response *res
 			Allocation:  nil,
 			SizeID:      size.ID,
 			PartitionID: partition.ID,
-			RackID:      requestPayload.RackID,
-			Hardware:    machineHardware,
+			// the RackID is calculated by the switch connections
+			// a metal-core instance might respond to pxe requests from all racks
+			//RackID:      requestPayload.RackID,
```
Gerrit91, line 597: Field should probably be deprecated / removed in the machine register request.

Gerrit91:
Works in test environment where we only have a single switch pair though.

(tested by powering off a waiting machine, fully wiping it from the database and then powering on again --> rackid was properly calculated)

cmd/metal-api/internal/service/machine-service.go
mwindower: Good point. I deprecated the field.

Gerrit91 (approved):

PR #183, opened 2021-04-07
Return unknownSize instead of nil if no matching size was found

majst01:
This will require to create a size with ID: "unknown" in the database to make this work

### DIFF
--- a/cmd/metal-api/internal/metal/size.go
+++ b/cmd/metal-api/internal/metal/size.go
@@ -95,10 +95,10 @@ nextsize:
 	}
 
 	if len(found) == 0 {
-		return nil, matchlog, NotFound("no size found for hardware (%s)", hardware.ReadableSpec())
+		return UnknownSize, matchlog, NotFound("no size found for hardware (%s)", hardware.ReadableSpec())
 	}
 	if len(found) > 1 {
-		return nil, matchlog, fmt.Errorf("%d sizes found for hardware (%s)", len(found), hardware.ReadableSpec())
+		return UnknownSize, matchlog, fmt.Errorf("%d sizes found for hardware (%s)", len(found), hardware.ReadableSpec())
 	}
 	return &found[0], []*SizeMatchingLog{matchedlog}, nil
 }
--- a/cmd/metal-api/internal/datastore/migrations/03_add_unknown_size.go
+++ b/cmd/metal-api/internal/datastore/migrations/03_add_unknown_size.go
@@ -1,0 +1,27 @@
+package migrations
+
+import (
+	r "gopkg.in/rethinkdb/rethinkdb-go.v6"
+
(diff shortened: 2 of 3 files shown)


cmd/metal-api/internal/metal/size.go
```diff
@@ -95,10 +95,10 @@ nextsize:
 	}
 
 	if len(found) == 0 {
-		return nil, matchlog, NotFound("no size found for hardware (%s)", hardware.ReadableSpec())
```
Gerrit91, line 98: There is also some defaulting applied in the `registerMachine` func. Is this really an error then?

cmd/metal-api/internal/metal/size.go
```diff
@@ -95,10 +95,10 @@ nextsize:
...
+		return UnknownSize, matchlog, NotFound("no size found for hardware (%s)", hardware.ReadableSpec())
 	}
 	if len(found) > 1 {
-		return nil, matchlog, fmt.Errorf("%d sizes found for hardware (%s)", len(found), hardware.ReadableSpec())
```
Gerrit91, line 101: I think in this case (which should actually never happen because we prevent overlapping), it should really just fail.

cmd/metal-api/internal/metal/size.go
majst01: No actually a "nil" size is returned

cmd/metal-api/internal/metal/size.go
majst01: It fails, what else do you mean with failing ?

cmd/metal-api/internal/metal/size.go
majst01: I think this is because there is no "unknown" size in the database

Gerrit91: Mh... I just think that returning an error + some ID is a little confusing for the caller. This is the caller:

```
	machineHardware := v1.NewMetalMachineHardware(&requestPayload.Hardware)
	size, _, err := r.ds.FromHardware(machineHardware)
	if err != nil {
		size = metal.UnknownSize
		utils.Logger(request).Sugar().Errorw("no size found for hardware, defaulting to unknown size", "hardware", machineHardware, "error", err)
	}
``` 

To me it feels better if we continue returning `nil` and in the code fragment above from `registerMachine` checks if the error was `NotFound` and then use the default there. In case it's not the `NotFound` error, the registration should fail and return 422 to the metal-hammer (undefined behavior). Should also help to explain my comment beneath a little better.

Gerrit91 (commented):
We had the `UnknownSize` transparent for the end-user in the past. This will change with this PR because it will then start to appear in `image ls`, I guess. We should probably add a description to the `UnknownSize`, which explains that it's a special entity used as a bucket for unmatched machines.

PR #188, opened 2021-05-05
Do not return consolepassword in machine describe anymore

### LINKED ISSUE #187: Remove consolepassword from machine describe response, add another API endpoint
In order to be able to specify more granular who is able to get the consolepassword and log access to it to a central location for external auditing. Additionally a reason must be given when requesting the consolepassword.

With this addition a log forwarder should be installed in the control-plane to ship logs of the metal-api to the SIEM

majst01:
add extra endpoint which is audit-able. Closes #187 

@droid42 any chance to gather user information in metal-api without mep-4 in place ?

Related PRs:

- https://github.com/metal-stack/metalctl/pull/87
- https://github.com/metal-stack/metal-go/pull/62

/cc @chbmuc


cmd/metal-api/main.go
```diff
@@ -710,7 +712,7 @@ func initRestServices(withauth bool) *restfulspec.Config {
 	restful.DefaultContainer.Filter(metrics.RestfulMetrics)
 
 	if withauth {
-		restful.DefaultContainer.Filter(rest.UserAuth(initAuth(lg.Sugar())))
+		restful.DefaultContainer.Filter(rest.UserAuth(userGetter))
```
majst01, line 715: What happens if main is called withoutAuth ?

Gerrit91: I think this is really just a branch for making the spec generation work. As soon as you get the spec generation and this feature working, everything should be at the right place.

majst01: make spec does not work, metal-api.json is empty, this should also be checked in ci that no changes to metal-api.json are created and fails if so.

majst01: added a CI job which ensures spec was generated and commited.

majst01: does not work

Gerrit91:
Can we prepare a metalctl PR for this new endpoint? That's somehow stopping me from trying out the PR in the mini-lab. Also small unit test would be neat.

cmd/metal-api/main.go
majst01: works now

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -440,6 +454,47 @@ func (r machineResource) findMachine(request *restful.Request, response *restful
...
+		return
+	}
+
+	m, err := r.ds.FindMachineByID(requestPayload.ID)
+	if checkError(request, response, utils.CurrentFuncName(), err) {
+		return
+	}
+
+	consolePassword := ""
+	if m.Allocation != nil && m.Allocation.ConsolePassword != "" {
```
Gerrit91, line 480: We should return an error when requesting a console password for an unallocated machine, otherwise a user retrieves `password:` as a result, which can be misleading.

majst01: fixed

Gerrit91 (approved):
Tried out successfully the PR in the mini-lab:

- ✅ Console password was removed from machine responses (list and describe)
- ✅ `metalctl machine consolepassword` gives back proper password
- ✅ Info log line when console password was requested is present
- ✅ No password returned without machine allocation

Gets my approval. However, one last thing to decide (from my side) is whether we want to make the validation of the reason optional through a metal-api CLI flag or something. Because for most users from the community giving a reason should not be required.

majst01:
I would open a new Issue which will make this configurable and implement that in a future release. WDYT ?

Gerrit91:
OK for me.

majst01:
implemented a configurable reason length which defaults to 0

Gerrit91 (approved):

PR #189, opened 2021-05-05
fix for duplicate ip allocation

### LINKED ISSUE #169: Investigate bug: Two machines in private network can end up having same IP addresses

majst01:
fixes #169

### DIFF
--- a/cmd/metal-api/internal/service/machine-service_allocation_test.go
+++ b/cmd/metal-api/internal/service/machine-service_allocation_test.go
@@ -1,0 +1,169 @@
+package service
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/emicklei/go-restful/v3"
+	goipam "github.com/metal-stack/go-ipam"
+	mdmv1 "github.com/metal-stack/masterdata-api/api/v1"
+	mdmv1mock "github.com/metal-stack/masterdata-api/api/v1/mocks"
+	mdm "github.com/metal-stack/masterdata-api/pkg/client"
+	"github.com/metal-stack/metal-api/cmd/metal-api/internal/datastore"
+	"github.com/metal-stack/metal-api/cmd/metal-api/internal/grpc"
+	"github.com/metal-stack/metal-api/cmd/metal-api/internal/ipam"
+	"github.com/metal-stack/metal-api/cmd/metal-api/internal/metal"
+	v1 "github.com/metal-stack/metal-api/cmd/metal-api/internal/service/v1"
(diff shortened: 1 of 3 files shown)


cmd/metal-api/internal/service/machine-service.go
```diff
@@ -944,7 +945,21 @@ func allocateMachine(logger *zap.SugaredLogger, ds *datastore.RethinkStore, ipam
 		}
 	}
 
-	machineCandidate, err := findMachineCandidate(ds, allocationSpec)
+	var machineCandidate *metal.Machine
+
+	err = retry.Do(
+		func() error {
+			var err2 error
+			machineCandidate, err2 = findMachineCandidate(ds, allocationSpec)
```
Gerrit91, line 953: Unfortunately, there is one more complication with the new approach.

After `findMachineCandidate` the `preAllocation` is set to `true`. It can then happen that the `allocateMachine` fails right after finding the machine candidate, for example in in `gatherNetworks`. The pre-allocation is then never set back to false, which makes the machine unusable forever. The rollback mechanism needs to be triggered even after `findMachineCandidate` now.

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -1337,13 +1353,15 @@ func makeMachineNetwork(ds *datastore.RethinkStore, ipamer ipam.IPAMer, allocati
 			ProjectID:        allocationSpec.ProjectID,
 		}
 		ip.AddMachineId(allocationSpec.UUID)
+		ds.Infow("create ip", "machine", allocationSpec.UUID, "ip", ipAddress)
```
Gerrit91, line 1356: I think this was only for debugging? We can clean it up.

cmd/metal-api/internal/service/machine-service_allocation_test.go
```diff
@@ -0,0 +1,224 @@
...
+	require.NoError(t, err)
+	err = rs.CreateNetwork(&metal.Network{Base: metal.Base{ID: "private"}, ParentNetworkID: "super", ProjectID: "pr1", PartitionID: "p1", Prefixes: metal.Prefixes{*privateNetwork}})
+	require.NoError(t, err)
+	err = rs.CreatePartition(&metal.Partition{Base: metal.Base{ID: "p1"}})
+	require.NoError(t, err)
+	err = rs.CreateSize(&metal.Size{Base: metal.Base{ID: "s1"}})
+	require.NoError(t, err)
+}
+
+type NopPublisher struct {
```
Gerrit91, line 213: I created a `common_test.go`, we can move it there.

Gerrit91 (approved):

PR #129, opened 2020-10-29
Add switch VRF update to async machine release mechanism.

Gerrit91:
This PR aims to mitigate errors like this:

```
...
2020-10-27 13:45:12	{"level":"info","time":"2020-10-27T12:45:12.651Z","caller":"service/machine-service.go:1574","message":"set VRF at switch","machineID":"262fa000-a9de-11e9-8000-ac1f6bd38d80"}
2020-10-27 13:45:13	{"level":"error","time":"2020-10-27T12:45:13.857Z","caller":"service/machine-service.go:1577","message":"cannot delete vrf switches","machineID":"262fa000-a9de-11e9-8000-ac1f6bd38d80","error":"cannot update switch (stg-kkw701-r01leaf01): the entity was changed from another, please retry"}
2020-10-27 13:45:13	{"level":"error","time":"2020-10-27T12:45:13.858Z","caller":"service/service.go:90","message":"service error","rqid":"a7f585fa3b7fa156173d206391049a60","remoteaddr":"10.20.2.12","method":"DELETE","uri":"/metal/v1/machine/262fa000-a9de-11e9-8000-ac1f6bd38d80/free","route":"/metal/v1/machine/{id}/free","useremail":"metal-admin@metal-stack.io","operation":"freeMachine","status":422,"error":"cannot delete vrf switches: cannot update switch (stg-kkw701-r01leaf01): the entity was changed from another, please retry","service-caller":"machine-service.go:1450","resp":"cannot delete vrf switches: cannot update switch (stg-kkw701-r01leaf01): the entity was changed from another, please retry (422)"}
2020-10-27 13:45:13	{"level":"error","time":"2020-10-27T12:45:13.858Z","caller":"rest/middleware.go:87","message":"Rest Call","rqid":"a7f585fa3b7fa156173d206391049a60","remoteaddr":"10.20.2.12","method":"DELETE","uri":"/metal/v1/machine/262fa000-a9de-11e9-8000-ac1f6bd38d80/free","route":"/metal/v1/machine/{id}/free","useremail":"metal-admin@metal-stack.io","rqid":"a7f585fa3b7fa156173d206391049a60","remoteaddr":"10.20.2.12","method":"DELETE","uri":"/metal/v1/machine/262fa000-a9de-11e9-8000-ac1f6bd38d80/free","route":"/metal/v1/machine/{id}/free","status":422,"content-length":160,"duration":1.399507635}
...
```

PR #192, opened 2021-05-09
free all allocated machines afterwards

majst01:


Gerrit91 (approved):
Do any errors occur without the retry?

I think the test addition is good, but we have to keep in mind that there still needs to be another test for freeing the machines that includes switch entities, because quite apparently there is another concurrency issue lying around, which is described here:

https://github.com/metal-stack/metal-api/pull/129

majst01:
Raised #193 to remember that

PR #194, opened 2021-05-31
Optionals

Gerrit91:
Declare optional values properly for swagger spec.
Python clients will error otherwise when null is retrieved.


majst01:
This requires adoptions on the metal-hammer as well

Gerrit91:
Yeah. I am sorry. The python client is picky when it comes to nullable-Fields because it validates the spec for incoming and outgoing requests. The current mep-8 implementation breaks machine responses for Python, listing filesystemlayouts doesn't work either.

majst01 (approved):

Issue #139, opened 2020-11-17
Store CreatorName in machine entities


Gerrit91:
I would assume this should be a very basic feature which helps us to identify who created a machine to a later point in time.

Gerrit91:
It would also be a good idea to store the name of the user who set a lock or a reserve state on a machine.

PR #197, opened 2021-06-18
added machine creator email info

GrigoriyMikhalkin:
Solves https://github.com/metal-stack/metal-api/issues/139

### DIFF
--- a/cmd/metal-api/internal/service/machine-service.go
+++ b/cmd/metal-api/internal/service/machine-service.go
@@ -53,6 +53,7 @@ type machineResource struct {
 
 // machineAllocationSpec is a specification for a machine allocation
 type machineAllocationSpec struct {
+	Creator            string
 	UUID               string
 	Name               string
 	Description        string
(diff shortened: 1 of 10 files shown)


cmd/metal-api/internal/service/machine-service.go
```diff
@@ -938,7 +939,13 @@ func (r machineResource) allocateMachine(request *restful.Request, response *res
 		}
 	}
 
+	user, err := r.userGetter.User(request.Request)
+	if checkError(request, response, utils.CurrentFuncName(), err) {
+		return
+	}
+
 	spec := machineAllocationSpec{
+		Creator:            user.EMail,
```
Gerrit91, line 948: The firewall service also creates such an allocation spec. The value has to be passed there as well.

Also, there is a function `validateAllocationSpec` which you can use to ensure that this field was properly passed through to the machine allocation logic.

GrigoriyMikhalkin: done

Gerrit91 (approved):

Gerrit91:
Could you please also create a PR in metal-go with the new swagger spec?

PR #200, opened 2021-07-02
Shorten BGP filter

### LINKED ISSUE #199: BGP filter is too wide
It contains private addresses of other partitions with /32 notation.
This yields in utterly long FRR route-maps.

mwindower:
fixes #199 

TODO:

- [ ] test with DMZ and Storage Setup in Test Env

### DIFF
--- a/cmd/metal-api/internal/metal/network.go
+++ b/cmd/metal-api/internal/metal/network.go
@@ -115,17 +115,7 @@ func (n *Network) FindPrefix(cidr string) *Prefix {
 
 // ContainsIP checks whether the given ip is included in the networks prefixes
 func (n *MachineNetwork) ContainsIP(ip string) bool {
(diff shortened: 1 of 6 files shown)


mwindower:
TODO: DMZ setup has to be considered

mwennrich (approved):

majst01:
This is not a fix, it just tries to shorten the route-maps in place and reduce therefore the probability for long running reloads which in turn sometimes lead to lost routes. The real fix is adding this fix to frr-reload.py on the leafs: https://github.com/FRRouting/frr/commit/6293c2181a3c58046f94496391d14a921a7fd892

majst01:
We should keep this PR open for reference, i still think its useful, we should discuss and test this at a later stage

majst01:
Simpler implementation was found with #559

PR #201, opened 2021-07-12
remove ioutil and errors, update dependencies

majst01:
- replace ioutil with newer funcs
- replace old errors
- update dependencies
- test new metal-stack/security

TODO:
- [x] test in test environment


cmd/metal-api/internal/metal/errors.go
```diff
@@ -13,30 +15,68 @@ var (
...
 // IsInternal checks if an error is a Internal error.
 func IsInternal(e error) bool {
-	return errors.Is(errors.Cause(e), errInternal)
+	return errors.Is(e, errInternal)
+}
+
+// wrapf returns an error annotating err with a stack trace
+// at the point Wrapf is called, and the format specifier.
+// If err is nil, Wrapf returns nil.
+func wrapf(err error, format string, args ...interface{}) error {
```
Gerrit91, line 49: This is the only part which makes me hesitate. Function is obviously taken from https://github.com/pkg/errors/blob/master/errors.go#L201, but is it really an improvement to copy it?

majst01: Agreed, i will revert this particular change, test are failing because of this i guess

majst01: reverted

Gerrit91:
Maybe needs protoc generation with 1.27.1? 

https://github.com/metal-stack/builder/pull/27

Gerrit91 (approved):

majst01:
i keep it running in test for a day or two before merging

majst01:
already included in #202

Issue #24, opened 2020-03-13
Add field(s) to clarify the function of the machine in the machine struct (regular machine, firewall, ...)


majst01:
I want to propose a field (boolean flags or a type field) to be added to the general machine struct in order to indicate whether a machine is a regular machine or a firewall (or something else for the future).

This would have multiple benefits:

- We get rid off the IsFirewall function (api/internal/metal/machine.go#L63-78), which is inefficient (because you have to pass all the images) and tends to be imprecise
- We can return this information directly to the clients as well within the machine list call. This has the advantage that for example the ansible dynamic inventory can derive the machine type pretty easily, which currently can only be determined by using multiple calls or the `image`-Features (see 
 ansible-common/blob/ips-and-networks/inventory/metal/metal.py#L99-109)

We can pass the flag or field directly in the firewall and machine create endpoints into the machine allocation spec, so this would be pretty easy to implement, I guess.

PR #211, opened 2021-08-10
Introduce allocation roles.

Gerrit91:
Closes https://github.com/metal-stack/metal-api/issues/24

### DIFF
--- a/cmd/metal-api/internal/datastore/machine.go
+++ b/cmd/metal-api/internal/datastore/machine.go
@@ -22,11 +22,12 @@ type MachineSearchQuery struct {
 	Tags        []string `json:"tags" optional:"true"`
 
 	// allocation
-	AllocationName      *string `json:"allocation_name" optional:"true"`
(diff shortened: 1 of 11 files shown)


Gerrit91:
Tried successfully in the mini-lab.

majst01 (approved):
Should we add constants in metal-lib for these 2 new introduced roles ?

Gerrit91:
In metal-lib? What do you mean?

majst01:
I mean for consumers of metal-go to be able to use this feature as well instead of comparing strings

Gerrit91:
I got you now. I added `enum` definitions to the swagger spec, such that the generated client now contains constants for the types as well.

majst01:
Glad to see the `enum` feature works now in swagger spec generation, we have plenty of other places where this would be very useful

Gerrit91:
I think this explicit annotation works since a long time, we already have it in many places. But automatic detection does not work for some reason. :/

majst01:
Migration must be tested in real world also

Issue #73, opened 2020-05-25
metalctl lies about network property `vrfshared`


mreiger:
When displaying a network with property `    vrfshared: true` with `metalctl network list --id=<network-id> -o yaml` it falsely claims `vrfshared` is set to `false`:

```
$ metalctl network list --id=<network-id> -o yaml
- changed: "2020-05-08T05:37:46.487Z"
  created: "2020-03-10T14:22:21.823Z"
  description: ""
  destinationprefixes:
    - <...>
  id: <network-id>
  labels: {}
  name: <network name>
  nat: true
  parentnetworkid: null
  partitionid: ""
  prefixes:
    - <network prefix>
  privatesuper: false
  projectid: ""
  underlay: false
  usage:
      availableips: 16
      availableprefixes: 0
      usedips: 15
      usedprefixes: 0
  vrf: <vrf number>
  vrfshared: false
```

majst01:
I think we can only decide between:
- persist in the metal.Network entity
- remove the property from the response

As this is only a flag for network creation and not used later in any case, i strongly vote removing this property from the response, and dont show it in the describe output. 
Removing from the response is also more difficult, because we need to modify/split the NetworkImmutable struct therefore.

PR #212, opened 2021-08-10
Remove vrfshared from network response.

Gerrit91:
Closes https://github.com/metal-stack/metal-api/issues/73.


Gerrit91:
Does this have any side effects I am not aware of? Solution looks so simple to me and our discussion in the past was fairly long. :thinking:

PR #219, opened 2021-08-25
Add filter options for partition capacity requests.

Gerrit91:


cmd/metal-api/internal/service/partition-service.go
```diff
@@ -103,11 +103,23 @@ func (r partitionResource) webService() *restful.WebService {
 		Returns(http.StatusConflict, "Conflict", httperrors.HTTPErrorResponse{}).
 		DefaultReturns("Error", httperrors.HTTPErrorResponse{}))
 
+	// Deprecated, can be removed in the future
```
majst01, line 106: as this is only used by "us" im fine removing it immediately.

Or is there a place in the cloud-api where we check for available machines ?

cmd/metal-api/internal/service/partition-service.go
```diff
@@ -290,29 +302,69 @@ func (r partitionResource) updatePartition(request *restful.Request, response *r
 	}
 }
 
+func (r partitionResource) partitionCapacityCompat(request *restful.Request, response *restful.Response) {
```
majst01, line 305: same

majst01 (changes requested):

cmd/metal-api/internal/service/partition-service.go
Gerrit91: Main reason for this decision to me was to have old metalctl versions still working for a certain time. In cloud-api we also call PartitionCapacity.

majst01 (approved):

PR #232, opened 2021-11-14
size image constraints types and validation

majst01:
Introduce SizeImageConstraints

SizeImageConstraint expresses optional restrictions for specific size to image combinations
this might be required if the support for a specific hardware in a given size is only supported
with a newer version of the image.

If the size in question is not found, no restrictions apply.
If the image in question is not found, no restrictions apply as well.
If the image in question is found, but does not match the given expression, machine creation must be forbidden.

Important:
API change, {machine|firewall}Allocate has now a try=true|false queryparameter

related PRs:

- [ ] https://github.com/metal-stack/metalctl/pull/122
- [ ] https://github.com/metal-stack/metal-go/pull/76


cmd/metal-api/internal/service/v1/firewall.go
```diff
@@ -2,6 +2,7 @@ package v1
 
 type FirewallCreateRequest struct {
 	MachineAllocateRequest
+	// FIXME remove this property, is nowhere used
```
majst01, line 5: Deprecate

cmd/metal-api/internal/service/firewall-service.go
```diff
@@ -172,84 +182,72 @@ func (r firewallResource) listFirewalls(request *restful.Request, response *rest
...
+	size := spec.Size
+	image := spec.Image
+
+	sic, err := r.ds.FindSizeImageConstraint(size.ID)
+	if checkError(request, response, utils.CurrentFuncName(), err) {
+		return
 	}
-	if ha {
-		if checkError(request, response, utils.CurrentFuncName(), errors.New("highly-available firewall not supported for the time being")) {
+	if sic != nil {
```
majst01, line 213: return if nil

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -915,6 +924,61 @@ func updateFru(m *metal.Machine, fru *v1.MachineFru) {
...
+	}
+	size := spec.Size
+	image := spec.Image
+
+	sic, err := r.ds.FindSizeImageConstraint(spec.Size.ID)
+	if checkError(request, response, utils.CurrentFuncName(), err) {
+		return
+	}
+
+	if sic == nil {
```
majst01, line 956: sic match should only be done if not nil, some in firewall-service

mwennrich (approved):

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -947,6 +975,26 @@ func (r machineResource) allocateMachine(request *restful.Request, response *res
...
+func isSizeAndImageCompatible(ds *datastore.RethinkStore, size metal.Size, image metal.Image) error {
+	sic, err := ds.FindSizeImageConstraint(size.ID)
+	if err != nil && !metal.IsNotFound(err) {
+		return err
+	}
+	if sic == nil {
+		return nil
+	}
+
+	ok, err := sic.Matches(size, image)
```
majst01, line 987: ok is useless

cmd/metal-api/internal/service/sizeimageconstraint-service.go
```diff
@@ -0,0 +1,211 @@
...
+			ds: ds,
+		},
+	}
+	return r.webService()
+}
+
+func (r sizeImageConstraintResource) webService() *restful.WebService {
+	ws := new(restful.WebService)
+	ws.
+		Path(BasePath + "v1/sizeimageconstraint").
```
Gerrit91, line 35: We prefer dash-case for our API endpoints.

cmd/metal-api/internal/service/firewall-service.go
```diff
@@ -107,6 +108,7 @@ func (r firewallResource) webService() *restful.WebService {
 		Operation("allocateFirewall").
 		Doc("allocate a firewall").
 		Metadata(restfulspec.KeyOpenAPITags, tags).
+		Param(ws.QueryParameter("try", "try allocation before actually doing so to get informed early about possible incompatible allocation parameters").DataType("bool").DefaultValue("false")).
```
Gerrit91, line 111: Again thinking about this parameter I've come to the conclusion that it's a bit clunky because testing the size image constraint now requires a lot of fields from the caller (see cloud-api trying to imitate future GEPM allocation requests).

Maybe it's better to add an individual `try` endpoint to the imagesizeconstraint service, which only requires a size id and an image id. The validation code called by this endpoint can be shared with and also be called from `allocateMachine`.

Gerrit91:
Integration tests need fixing

majst01:
> Integration tests need fixing

on its way

Makefile
```diff
@@ -35,3 +35,7 @@ mini-lab-push:
 	kind --name metal-control-plane load docker-image metalstack/metal-api:latest
 	kubectl --kubeconfig=$(MINI_LAB_KUBECONFIG) patch deployments.apps -n metal-control-plane metal-api --patch='{"spec":{"template":{"spec":{"containers":[{"name": "metal-api","imagePullPolicy":"IfNotPresent","image":"metalstack/metal-api:latest"}]}}}}'
 	kubectl --kubeconfig=$(MINI_LAB_KUBECONFIG) delete pod -n metal-control-plane -l app=metal-api
+
+.PHONY: integration
+integration:
+	go test -tags=integration -timeout 600s -p 1 ./...
```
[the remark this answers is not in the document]
majst01: removed

Gerrit91 (approved):

PR #235, opened 2021-11-24
refactor allocation spec creation

majst01:
With this PR the duplicated code to fill the internal struct allocationspec was refactored to be reusable for machine and firewall allocation. Also the FirewallCreateRequest.HA field was never used and deprecated here.

This PR is a preparation for #232 to make the review of this PR easier.

### DIFF
--- a/cmd/metal-api/internal/service/machine-service.go
+++ b/cmd/metal-api/internal/service/machine-service.go
@@ -60,6 +60,7 @@ type machineAllocationSpec struct {
 	Hostname           string
 	ProjectID          string
 	PartitionID        string
+	Machine            *metal.Machine
 	Size               *metal.Size
 	Image              *metal.Image
 	FilesystemLayoutID *string
@@ -987,6 +988,7 @@ func createMachineAllocationSpec(ds *datastore.RethinkStore, requestPayload v1.M
 
 	partitionID := requestPayload.PartitionID
 	sizeID := requestPayload.SizeID
+	var m *metal.Machine
 	// Allocation of a specific machine is requested, therefor size and partition are not given, fetch them
 	if uuid != "" {
 		m, err := ds.FindMachineByID(uuid)
@@ -1010,6 +1012,7 @@ func createMachineAllocationSpec(ds *datastore.RethinkStore, requestPayload v1.M
 		Hostname:    hostname,
 		ProjectID:   requestPayload.ProjectID,
 		PartitionID: partitionID,
+		Machine:     m,
 		Size:        size,
 		Image:       image,
 		SSHPubKeys:  requestPayload.SSHPubKeys,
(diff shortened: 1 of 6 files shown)


cmd/metal-api/internal/service/machine-service.go
```diff
@@ -942,54 +967,58 @@ func (r machineResource) allocateMachine(request *restful.Request, response *res
 	if requestPayload.UserData != nil {
 		userdata = *requestPayload.UserData
 	}
+	if requestPayload.Networks != nil && len(requestPayload.Networks) <= 0 {
```
majst01, line 970: We should check for nil networks as well !

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -942,54 +967,58 @@ func (r machineResource) allocateMachine(request *restful.Request, response *res
...
-		Role:               metal.RoleMachine,
-	}
-
-	m, err := allocateMachine(utils.Logger(request).Sugar(), r.ds, r.ipamer, &spec, r.mdc, r.actor, r.grpcServer)
-	if checkError(request, response, utils.CurrentFuncName(), err) {
-		utils.Logger(request).Sugar().Errorw("machine allocation went wrong", "error", err)
-		return
+	partitionID := requestPayload.PartitionID
+	sizeID := requestPayload.SizeID
+	// Allocation of a specific machine is requested, therefor size and partition are not given, fetch them
```
Gerrit91, line 990:
```suggestion
	// Allocation of a specific machine is requested, therefore size and partition are not given, fetch them
```

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -942,54 +967,58 @@ func (r machineResource) allocateMachine(request *restful.Request, response *res
...
+	if uuid != "" {
+		m, err := ds.FindMachineByID(uuid)
+		if err != nil {
+			return nil, fmt.Errorf("uuid given but no machine found with uuid:%s err:%w", uuid, err)
+		}
```
Gerrit91, line 995: Introduces an additional backend call, which I think gets repeated in `findMachineCandidate`. Maybe it makes sense to store the pointer to this machine entity in the allocation spec and we can get rid off UUID plus the additional backend call.

Gerrit91 (commented):
Thanks for putting this into a separate PR. Makes it much better to understand. :)

cmd/metal-api/internal/service/machine-service.go
majst01: Good idea, done

majst01:
Unrelated but its possible to allocate a firewall which leads to a failure during installation:

```
{"level":"panic","ts":1637767309.1289444,"caller":"cmd/root.go:93","msg":"expectation at least one public network (private: false, underlay: false) is present failed","stacktrace":"github.com/metal-stack/metal-networker/cmd.configure\n\tgithu
b.com/metal-stack/metal-networker/cmd/root.go:93\ngithub.com/metal-stack/metal-networker/cmd.glob..func1\n\tgithub.com/metal-stack/metal-networker/cmd/firewall.go:24\ngithub.com/spf13/cobra.(*Command).execute\n\tgithub.com/spf13/cobra@v1.1.3/
command.go:852\ngithub.com/spf13/cobra.(*Command).ExecuteC\n\tgithub.com/spf13/cobra@v1.1.3/command.go:960\ngithub.com/spf13/cobra.(*Command).Execute\n\tgithub.com/spf13/cobra@v1.1.3/command.go:897\ngithub.com/metal-stack/metal-networker/cmd.
Execute\n\tgithub.com/metal-stack/metal-networker/cmd/root.go:43\nmain.main\n\tgithub.com/metal-stack/metal-networker/main.go:6\nruntime.main\n\truntime/proc.go:225"}                                                                            
panic: expectation at least one public network (private: false, underlay: false) is present failed  
```

Gerrit91 (approved):

PR #237, opened 2021-11-25
Update retry-go to v4 which is go modules compatible

majst01:


cmd/metal-api/internal/service/machine-service_allocation_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build integration
 // +build integration
```
Gerrit91, line 2:
```suggestion
```

Gerrit91 (approved):

PR #241, opened 2021-12-03
Build with buildx to speed up docker image creation

majst01:
Tried, but not faster.


majst01:
Should be migrated to docker/metadata-action and  docker/build-push-action

Issue #242, opened 2021-12-03
Store machine Events in a dedicated rethinkdb


majst01:

vknabel:
Closed due to age. If still needed please reopen.

PR #247, opened 2021-12-14
Store Events in separate database

majst01:
related to #242

Breaking: event endpoints moved to /v1/events

### DIFF
--- a/cmd/metal-api/internal/datastore/event.go
+++ b/cmd/metal-api/internal/datastore/event.go
@@ -1,7 +1,11 @@
 package datastore
 
 import (
+	"fmt"
+	"time"
+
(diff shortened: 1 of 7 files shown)


Gerrit91:
Why does it need to API breaking? Can't we just write to an alternative datastore for both endpoints if enabled?

majst01:
> Why does it need to API breaking? Can't we just write to an alternative datastore for both endpoints if enabled?

This not my final intention, i would like to have some sort of alias from the old `/v1/machine/event` route to `/v1/event...` route.

vknabel:
Closed as PR is too old. Please rebase and submit a new PR if still needed.

PR #262, opened 2022-03-17
Error out when referenced entities not found for switches and machines.

Gerrit91:
This fix is really urgent as it can prevent very bad things to happen, e.g. returning an empty switch response which causes metal-core to render empty FRR configurations or return half.empty machine responses, which can cause components like MCM to handle machines in a wrong way.

### DIFF
--- a/cmd/metal-api/internal/service/machine-service.go
+++ b/cmd/metal-api/internal/service/machine-service.go
@@ -448,7 +448,13 @@ func (r machineResource) listMachines(request *restful.Request, response *restfu
 	if checkError(request, response, utils.CurrentFuncName(), err) {
 		return
 	}
-	err = response.WriteHeaderAndEntity(http.StatusOK, makeMachineResponseList(ms, r.ds, utils.Logger(request).Sugar()))
+
+	resp, err := makeMachineResponseList(ms, r.ds)
+	if checkError(request, response, utils.CurrentFuncName(), err) {
+		return
+	}
+
+	err = response.WriteHeaderAndEntity(http.StatusOK, resp)
 	if err != nil {
 		zapup.MustRootLogger().Error("Failed to send response", zap.Error(err))
 		return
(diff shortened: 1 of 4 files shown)


cmd/metal-api/internal/service/machine-service.go
```diff
@@ -2316,13 +2390,19 @@ func machineHasIssues(m *v1.MachineResponse) bool {
...
 
-func makeMachineResponse(m *metal.Machine, ds *datastore.RethinkStore, logger *zap.SugaredLogger) *v1.MachineResponse {
-	s, p, i, ec := findMachineReferencedEntities(m, ds, logger)
-	return v1.NewMachineResponse(m, s, p, i, ec)
+func makeMachineResponse(m *metal.Machine, ds *datastore.RethinkStore) (*v1.MachineResponse, error) {
+	s, p, i, ec, err := findMachineReferencedEntities(m, ds)
+	if err != nil {
+		return nil, err
+	}
+	return v1.NewMachineResponse(m, s, p, i, ec), nil
```
majst01, line 2398: Check if v1.NewMachineResponse should return an Error as well, or if findMachineReferencedEntities ensures to return either error or none of the return values are non nil

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -2348,16 +2428,22 @@ func makeMachineResponseList(ms metal.Machines, ds *datastore.RethinkStore, logg
...
 
-func makeMachineIPMIResponse(m *metal.Machine, ds *datastore.RethinkStore, logger *zap.SugaredLogger) *v1.MachineIPMIResponse {
-	s, p, i, ec := findMachineReferencedEntities(m, ds, logger)
-	return v1.NewMachineIPMIResponse(m, s, p, i, ec)
+func makeMachineIPMIResponse(m *metal.Machine, ds *datastore.RethinkStore) (*v1.MachineIPMIResponse, error) {
+	s, p, i, ec, err := findMachineReferencedEntities(m, ds)
+	if err != nil {
+		return nil, err
+	}
+	return v1.NewMachineIPMIResponse(m, s, p, i, ec), nil
```
majst01, line 2439: same as above

cmd/metal-api/internal/service/switch-service.go
```diff
@@ -604,11 +639,14 @@ func connectMachineWithSwitches(ds *datastore.RethinkStore, m *metal.Machine) er
...
-	p, ips, machines := findSwitchReferencedEntites(s, ds, logger)
+func makeSwitchResponse(s *metal.Switch, ds *datastore.RethinkStore) (*v1.SwitchResponse, error) {
+	p, ips, machines, err := findSwitchReferencedEntites(s, ds)
+	if err != nil {
+		return nil, err
+	}
 	nics := makeSwitchNics(s, ips, machines)
 	cons := makeSwitchCons(s)
-	return v1.NewSwitchResponse(s, p, nics, cons)
+	return v1.NewSwitchResponse(s, p, nics, cons), nil
```
majst01, line 649: same as above

majst01 (approved):

PR #146, opened 2020-11-23
Resurrect machine without allocation that are phoning home and alive

muhittink:

PR #265, opened 2022-03-18
Introduce state machine (FSM) to track machine provisioning sequence

### LINKED ISSUE #264: Introduce state machine to track machine provisioning sequence
As mentioned in #166 and #167 there are some problems that can occur during machine provisioning, that are not yet being reported to the user, because handling all the potential scenarios would lead to a complex logic.
A state machine might be a good solution for this, as it can individually react to any sequence of events while still being easy to understand and extend.

iljarotar:
Using a fsm to detect an unexpected order of provisioning events might be helpful for handling cases like mentioned in  #167 and  #166

Closes #264

```ACTIONS_REQUIRED
The field `IncompleteProvisioningCycles` in the provisioning events container is now deprecated and replaced by two new bool flags `CrashLoop` and `MachineReclaimFailed`.
```


cmd/metal-api/internal/metal/provisioning.go
```diff
@@ -4,9 +4,56 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/looplab/fsm"
 	"go.uber.org/zap"
 )
 
+// FSM states
+const (
+	FsmStateInitial          string = "Initial State"
```
Gerrit91, line 13: Not sure why copying the states is necessary when they are already defined below?

cmd/metal-api/internal/metal/provisioning.go
```diff
@@ -4,9 +4,56 @@ import (
...
+	FsmStatePreparing        string = "Rreparing state"
+	FsmStateRegistering      string = "Registering state"
+	FsmStateWaiting          string = "Waiting state"
+	FsmStateInstalling       string = "Installing state"
+	FsmStateBootingNewKernel string = "Booting New Kernel state"
+	FsmStateProvisioned      string = "Provisioned state"
+	FsmStatePlannedReboot    string = "Planned Reboot state"
+	FsmStateCrashed          string = "Crashed state"
+	FsmStateAlive            string = "Alive state"
+	FsmStateResestFailCount  string = "Reset Fail Count state"
```
Gerrit91, line 24: Typo "Resest"

cmd/metal-api/internal/metal/provisioning.go
```diff
@@ -4,9 +4,56 @@ import (
...
+	FsmStateBootingNewKernel,
+	FsmStateProvisioned,
+	FsmStatePlannedReboot,
+	FsmStateCrashed,
+	FsmStateAlive,
+	FsmStateResestFailCount,
+}
+
+var events = []fsm.EventDesc{
+	{Name: string(ProvisioningEventPXEBooting), Src: []string{FsmStateInitial, FsmStatePlannedReboot, FsmStatePXEBooting, FsmStateCrashed, FsmStateProvisioned, FsmStateResestFailCount}, Dst: FsmStatePXEBooting},
```
Gerrit91, line 43: To improve readability I want to propose some line breaks here.

This way, we would also be able to comment some transitions that are not directly obvious, e.g. `// PXE Boot events practically occur multiple times in a row, which is considered fine`

cmd/metal-api/internal/metal/provisioning.go
```diff
@@ -4,9 +4,56 @@ import (
...
+	{Name: string(ProvisioningEventRegistering), Src: []string{FsmStatePreparing, FsmStateInitial, FsmStateResestFailCount}, Dst: FsmStateRegistering},
+	{Name: string(ProvisioningEventWaiting), Src: []string{FsmStateRegistering, FsmStateInitial, FsmStateResestFailCount}, Dst: FsmStateWaiting},
+	{Name: string(ProvisioningEventInstalling), Src: []string{FsmStateWaiting, FsmStateInitial, FsmStateResestFailCount}, Dst: FsmStateInstalling},
+	{Name: string(ProvisioningEventBootingNewKernel), Src: []string{FsmStateInstalling, FsmStateInitial, FsmStateResestFailCount}, Dst: FsmStateBootingNewKernel},
+	{Name: string(ProvisioningEventPhonedHome), Src: []string{FsmStateBootingNewKernel, FsmStateProvisioned, FsmStateInitial, FsmStateResestFailCount}, Dst: FsmStateProvisioned},
+	{Name: string(ProvisioningEventPlannedReboot), Src: allStates, Dst: FsmStatePlannedReboot},
+	{Name: string(ProvisioningEventCrashed), Src: []string{FsmStateInitial}, Dst: FsmStateCrashed},
+	{Name: string(ProvisioningEventResetFailCount), Src: allStates, Dst: FsmStateResestFailCount},
```
Gerrit91, line 52: Over time it turned out that the `ProvisioningEventResetFailCount` is not really used and also not so useful. From my side, you can remove this state from your solution.

cmd/metal-api/internal/metal/provisioning.go
```diff
@@ -4,9 +4,56 @@ import (
...
+}
+
+var callbacks = map[string]fsm.Callback{}
```
Gerrit91, line 55: I guess these will be implemented at a later stage? If so, you can imply your intent with `// TODO: to be implemented` or something like this.

cmd/metal-api/internal/metal/provisioning.go
```diff
@@ -100,42 +134,40 @@ func (p provisioningEventSequence) containsEvent(event ProvisioningEventType) bo
...
-	currentEvent := e.Event
-
-	expectedSuccessors, ok := expectedSuccessorEventMap[currentEvent]
-	if !ok {
-		log.Errorw("successor map does not contain an expected successor for event", "event", currentEvent)
-		return false
+func postEvent(currentFSM *fsm.FSM, event ProvisioningEventType) (*fsm.FSM, error) {
+	var nextFSM *fsm.FSM
+	err := currentFSM.Event(string(event))
+	if err != nil && err.Error() != "no transition" {
```
Gerrit91, line 140: Checking for specific errors can be done with `errors.Is` or `errors.As` (https://pkg.go.dev/errors#Is), string comparison is error-prone. The state machine library returns a typed `NoTransition` error.

cmd/metal-api/internal/metal/provisioning.go
```diff
@@ -100,42 +134,40 @@ func (p provisioningEventSequence) containsEvent(event ProvisioningEventType) bo
...
 }
 
-func (e *ProvisioningEvent) hasExpectedSuccessor(log *zap.SugaredLogger, actualSuccessor ProvisioningEventType) bool {
-	currentEvent := e.Event
-
-	expectedSuccessors, ok := expectedSuccessorEventMap[currentEvent]
-	if !ok {
-		log.Errorw("successor map does not contain an expected successor for event", "event", currentEvent)
-		return false
+func postEvent(currentFSM *fsm.FSM, event ProvisioningEventType) (*fsm.FSM, error) {
```
Gerrit91, line 137: You return `*fsm.FSM` but the caller does not use it. So I think there is no need to pass it around. I guess you want to make this function the place where you handle a new incoming machine event? Then maybe just pass the incoming event and use the signature `HandleMachineEvent(event ProvisioningEventType) error` or something? Later, we will probably also pass the machine entity, too, such that we can update the provisioning event container in the callback functions.

cmd/metal-api/internal/metal/provisioning.go
```diff
@@ -100,42 +134,40 @@ func (p provisioningEventSequence) containsEvent(event ProvisioningEventType) bo
...
 
 // CalculateIncompleteCycles calculates the number of events that occurred out of order. Can be used to determine if a machine is in an error loop.
 func (p *ProvisioningEventContainer) CalculateIncompleteCycles(log *zap.SugaredLogger) string {
+	fsm := newProvisioningFSM()
 	incompleteCycles := 0
-	atLeastOneTimeCompleted := true
-	var successor ProvisioningEventType
-	for i, event := range p.Events {
-		if successor != "" && !event.hasExpectedSuccessor(log, successor) {
+	var err error
```
Gerrit91, line 161: I don't think you need to take care so much of the "incomplete cycles"-part for now because it's just sitting on top of the rest of the functionality. We can probably discuss to replace it by a single "CrashLoop"-flag or something in a later iteration.

So I think, I would like if this function is just kept as it is and we clean it up in the end when we don't need it anymore.

cmd/metal-api/internal/metal/provisioning_test.go
```diff
@@ -196,65 +173,57 @@ func TestProvisioning_IncompleteCycles(t *testing.T) {
...
+			name: "TestFsm_CrashCycles Test 7",
 			eventContainer: ProvisioningEventContainer{
-				Events: SuccessfulEventCycleWithBadHistory,
+				Events: SuccessfulCycleWithBadHistory,
 			},
 			want: "0",
 		},
 		{
-			name: "TestProvisioning_IncompleteCycles Test 9",
+			name: "TestFsm_CrashCycles Test 8",
```
Gerrit91, line 224: General remark, not this line specifically: Just introduce a new test for your code, the `CalculateIncompleteCycles` is not so important for the beginning. In case it starts to fail, please just comment this test.

Gerrit91 (commented):
I left some comment remarks that I hope help. The CI build was failing due to an old flaky test that I think we should remove or fix anytime soon. Retrying helps for the time being.

We can also do a small meet for discussion or pair programming session to build the foundation of this PR when you are interested.

cmd/metal-api/internal/fsm/fsm.go
```diff
@@ -0,0 +1,304 @@
...
+	"strconv"
+	"time"
+
+	"github.com/looplab/fsm"
+	"github.com/metal-stack/metal-api/cmd/metal-api/internal/metal"
+	"go.uber.org/zap"
+)
+
+type EventContainer struct {
+	IncompleteCycles int
```
Gerrit91, line 15: Can be a simple flag `Crashloop`

cmd/metal-api/internal/fsm/fsm.go
```diff
@@ -0,0 +1,304 @@
...
+	"reflect"
+	"strconv"
+	"time"
+
+	"github.com/looplab/fsm"
+	"github.com/metal-stack/metal-api/cmd/metal-api/internal/metal"
+	"go.uber.org/zap"
+)
+
+type EventContainer struct {
```
Gerrit91, line 14: Consider re-using existing struct from metal package.

cmd/metal-api/internal/fsm/fsm.go
```diff
@@ -0,0 +1,304 @@
...
+	LastEventTime    *time.Time
+	Events           metal.ProvisioningEvents
+	Liveliness       metal.MachineLiveliness
+}
+
+const timeOutAfterPlannedReboot = time.Minute * 5
+
+var events = fsm.Events{
+	{
+		Name: string(metal.ProvisioningEventPXEBooting),
```
Gerrit91, line 26: Stringer with `.String()` call is probably a bit nicer to read.

cmd/metal-api/internal/fsm/fsm.go
```diff
@@ -0,0 +1,304 @@
...
+			string(metal.ProvisioningEventPlannedReboot),
+			string(metal.ProvisioningEventPhonedHome),
+		},
+		Dst: string(metal.ProvisioningEventPXEBooting),
+	},
+	{
+		Name: string(metal.ProvisioningEventPreparing),
+		Src: []string{
+			string(metal.ProvisioningEventPXEBooting),
+			string(metal.ProvisioningEventPlannedReboot),
```
Gerrit91, line 38: Can you add a comment why this is a valid src?

cmd/metal-api/internal/fsm/fsm.go
```diff
@@ -0,0 +1,304 @@
...
+		Name: string(metal.ProvisioningEventPlannedReboot),
+		Src: []string{
+			string(metal.ProvisioningEventPXEBooting),
+			string(metal.ProvisioningEventPreparing),
+			string(metal.ProvisioningEventRegistering),
+			string(metal.ProvisioningEventWaiting),
+			string(metal.ProvisioningEventInstalling),
+			string(metal.ProvisioningEventBootingNewKernel),
+			string(metal.ProvisioningEventPhonedHome),
+			string(metal.ProvisioningEventAlive),
```
Gerrit91, line 89: Can be removed.

cmd/metal-api/internal/fsm/fsm.go
```diff
@@ -0,0 +1,304 @@
...
+	container, event, _, err := parseCallbackArguments(e.Args)
+	if err != nil {
+		e.Err = err
+		return
+	}
+
+	if e.Src == string(metal.ProvisioningEventPlannedReboot) {
+		if event.Time.Sub(*container.LastEventTime) > timeOutAfterPlannedReboot {
+			container.Liveliness = metal.MachineLivelinessUnknown
+		}
```
Gerrit91, line 173: Is a bit different: The problematic state only comes up when the machine is being freed (reclaimed) and then boots back up into the user's OS emitting Phoned Home events again. Instead maybe introduce a new state `MachineReclaim` which is emitted during machine free.

cmd/metal-api/internal/fsm/fsm.go
```diff
@@ -0,0 +1,304 @@
...
+
+func handleAliveEvent(e *fsm.Event) {
+	container, _, log, err := parseCallbackArguments(e.Args)
+	if err != nil {
+		e.Err = err
+		return
+	}
+
+	log.Debugw("received provisioning alive event", "id", container.ID)
+	container.Liveliness = metal.MachineLivelinessAlive
```
Gerrit91, line 191: Also need to set last event time such that it will not flap when machine liveliness is evaluated.

cmd/metal-api/internal/fsm/fsm.go
```diff
@@ -0,0 +1,304 @@
...
+		IncompleteCycles: 0,
+		ID:               container.ID,
+		Liveliness:       container.Liveliness,
+		Events:           []metal.ProvisioningEvent{container.Events[0]},
+		LastEventTime:    container.LastEventTime,
+	}
+
+	var err error
+	var invalidEventError fsm.InvalidEventError
+	for i, e := range container.Events {
```
Gerrit91, line 241: When you use just the flag `Crashloop` there is no need to cycle through the event history anymore.

cmd/metal-api/internal/fsm/fsm.go
```diff
@@ -0,0 +1,306 @@
...
+			}
+		} else if reflect.TypeOf(err) == reflect.TypeOf(invalidEventError) {
+			container.Events = append(container.Events, *event)
+			container.LastEventTime = &(*event).Time
+			container.CrashLoop = true
+			err = nil
+		}
+	}
+
+	return container, err
```
Gerrit91, line 275: From style perspective it's better to avoid deeply nested code. You can achieve this with early returns, too:

```golang
	err := provisioningFSM.fsm.Event(event.Event.String(), provisioningFSM, event)
	if err == nil {
		return container, nil
	}

	if errors.As(err, fsm.NoTransitionError{}) {
		container.LastEventTime = &(*event).Time
		container.Liveliness = metal.MachineLivelinessAlive

		if event.Event == metal.ProvisioningEventAlive {
			log.Debugw("received provisioning alive event", "id", container.ID)
		} else if event.Event == metal.ProvisioningEventPhonedHome {
			log.Debugw("swallowing repeated phone home event", "id", container.ID)
		}

		return container, nil
	}

	if errors.As(err, &fsm.InvalidEventError{}) {
		container.Events = append(container.Events, *event)
		container.LastEventTime = &(*event).Time
		container.CrashLoop = true

		return container, nil
	}

	return nil, err
```

cmd/metal-api/internal/fsm/fsm.go
```diff
@@ -0,0 +1,303 @@
...
+	provisioningFSM := ProvisioningFSM{
+		fsm:       f,
+		container: container,
+	}
+
+	return &provisioningFSM
+}
+
+// HandleProvisioningEvent writes the ProvisioningEvent to the the ProvisioningEventContainer and checks if it is a valid event in the current state
+func HandleProvisioningEvent(event *metal.ProvisioningEvent, container *metal.ProvisioningEventContainer, log *zap.SugaredLogger) (*metal.ProvisioningEventContainer, error) {
```
Gerrit91, line 244: It's a good practice to place public functions before private functions.

cmd/metal-api/internal/fsm/fsm.go
```diff
@@ -0,0 +1,303 @@
...
+	if len(container.Events) == 0 {
+		container.Events = append(container.Events, *event)
+		container.LastEventTime = &(*event).Time
```
Gerrit91, line 247: Just `&event.Time`? You should not dereference here.

cmd/metal-api/internal/fsm/fsm.go
```diff
@@ -0,0 +1,303 @@
...
+		events,
+		callbacks,
+	)
+
+	provisioningFSM := ProvisioningFSM{
+		fsm:       f,
+		container: container,
+	}
+
+	return &provisioningFSM
```
Gerrit91, line 240: When returning this directly, you also do it inline:

```suggestion
	return &ProvisioningFSM{
		fsm:       f,
		container: container,
	}
```

cmd/metal-api/internal/fsm/fsm.go
```diff
@@ -0,0 +1,303 @@
...
+			container.LastEventTime = &(*event).Time
+			container.CrashLoop = true
+			err = nil
+		}
+	}
+
+	return container, err
+}
+
+func ProvisioningEventForMachine(log *zap.SugaredLogger, ec *metal.ProvisioningEventContainer, machineID, event, message string) *metal.ProvisioningEventContainer {
```
Gerrit91, line 278: I think it's better to pass the `metal.ProvisioningEvent` and not event and message strings. One less func arg and caller has the required information.

Also you probably don't need to pass the machineID as it should be contained in the provisioning event container. Let the caller pass it properly and don't allow it to be nil by returning an error.

cmd/metal-api/internal/fsm/fsm.go
```diff
@@ -0,0 +1,303 @@
...
+	now := time.Now()
+	ec.LastEventTime = &now
+
+	ev := metal.ProvisioningEvent{
+		Time:    now,
+		Event:   metal.ProvisioningEventType(event),
+		Message: message,
+	}
+
+	ec, err := HandleProvisioningEvent(&ev, ec, log)
```
Gerrit91, line 296: Does this still need to be public? It looks like from the outside you only need to call this function?

cmd/metal-api/internal/fsm/fsm.go
```diff
@@ -0,0 +1,303 @@
...
+	if err != nil {
+		log.Errorf("internal error while calculating provisioning event container for machine %v", machineID)
```
Gerrit91, line 298: Give back the error from this function and let the caller decide how to handle the issue.

cmd/metal-api/internal/fsm/fsm.go
```diff
@@ -0,0 +1,303 @@
...
+	}
+
+	ec.TrimEvents(metal.ProvisioningEventsInspectionLimit)
```
Gerrit91, line 301: Probably also something that the caller should do, who is responsible for persisting the container.

cmd/metal-api/internal/fsm/fsm.go
```diff
@@ -0,0 +1,303 @@
...
+
+func handleStateTransition(e *fsm.Event) {
+	provisioningFSM, event, err := parseCallbackArguments(e.Args)
+	if err != nil {
+		e.Err = err
+		return
+	}
+
+	provisioningFSM.container.Events = append(provisioningFSM.container.Events, *event)
+	provisioningFSM.container.LastEventTime = &(*event).Time
```
Gerrit91, line 165: `&event.Time`

cmd/metal-api/internal/fsm/fsm.go
```diff
@@ -0,0 +1,303 @@
...
+	"enter_" + metal.ProvisioningEventBootingNewKernel.String(): handleStateTransition,
+	"enter_" + metal.ProvisioningEventMachineReclaim.String():   handleStateTransition,
+
+	"enter_" + metal.ProvisioningEventPXEBooting.String():    resetFailedReclaim,
+	"enter_" + metal.ProvisioningEventPreparing.String():     resetFailedReclaim,
+	"enter_" + metal.ProvisioningEventPlannedReboot.String(): handlePlannedRebootEvent,
+	"enter_" + metal.ProvisioningEventPhonedHome.String():    handlePhonedHomeEvent,
+}
+
+func handleStateTransition(e *fsm.Event) {
```
Gerrit91, line 157: You can define the callbacks on `ProvisioningFSM` and define the callbacks inline when creating it. Then you don't need to pass it through the arguments.

Gerrit91 (changes requested):

majst01:
This looks already quite promising, a dream comes true !

cmd/metal-api/internal/fsm/fsm.go
```diff
@@ -0,0 +1,262 @@
...
+	{
+		Name: metal.ProvisioningEventAlive.String(),
+		Src: []string{
+			metal.ProvisioningEventBootingNewKernel.String(),
+		},
+		Dst: metal.ProvisioningEventBootingNewKernel.String(),
+	},
+}
+
+func ProvisioningEventForMachine(log *zap.SugaredLogger, ec *metal.ProvisioningEventContainer, event *metal.ProvisioningEvent) (*metal.ProvisioningEventContainer, error) {
```
Gerrit91, line 144: Can we add a comment to this func and generally explain the idea of calling it? Maybe even rename this function to `HandleProvisioningEvent`.

And for documentation something like: 

```golang
// HandleProvisioningEvent can be called to determine whether the given incoming event follows an expected lifecycle of a machine considering the event history of the given provisioning event container.
//
// The function returns a new provisioning event container that can then be persisted in the database. If an error is returned, the incoming event is not supposed to be persisted in the database.
//
// Among other things, this function can detect crash loops or other irregularities within a machine lifecycle and enriches the returned provisioning event container with this information. 
```

cmd/metal-api/internal/fsm/fsm_test.go
```diff
@@ -0,0 +1,270 @@
...
+			wantCrashLoop:      false,
+			wantFailedReclaim:  true,
+			wantLiveliness:     metal.MachineLivelinessAlive,
+			wantNumberOfEvents: 1,
+			wantLastEventTime:  now.Add(time.Minute * 10),
+			wantLastEvent:      metal.ProvisioningEventMachineReclaim.String(),
+		},
+	}
+	for _, tt := range tests {
+		log := zap.NewExample().Sugar()
```
Gerrit91, line 237: Better to use `zaptest.NewLogger(t).Sugar()`.

cmd/metal-api/internal/fsm/fsm_test.go
```diff
@@ -0,0 +1,270 @@
...
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			_, err := handleProvisioningEvent(tt.event, tt.container, log)
+			if diff := cmp.Diff(tt.wantErr, err); diff != "" {
+				t.Errorf("HandleProvisioningEvent() diff = %s", diff)
+			}
+
+			if tt.container.CrashLoop != tt.wantCrashLoop {
```
Gerrit91, line 245: Why do you compare against `tt.container.CrashLoop` and not the event container returned by `handleProvisioningEvent`?

You can also do this check in one-line: `assert.Equal(t, tt.wantCrashLoop, got.CrashLoop)`

Gerrit91 (changes requested):
Left some remarks on the code.

Also I think I would like to see (even it could be a breaking change) a `func (e *ProvisioningEventcontainer) Validate() error` in the `metal` package, which can verify that the event container is good. Checks can include:

- Verify that last event time is newer or equal than newest event in the event list
- Check that all events in the list are chronologically sorted

We can then validate an event container before storing it to the database, to prevent inconsistent data to be stored in the database.

Gerrit91:
Progress is great by the way! It starts to look like something. 😍

cmd/metal-api/internal/fsm/fsm.go
```diff
@@ -0,0 +1,231 @@
...
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/looplab/fsm"
+	"github.com/metal-stack/metal-api/cmd/metal-api/internal/metal"
+	"go.uber.org/zap"
+)
+
+const timeOutAfterMachineReclaim = time.Minute * 5
```
Gerrit91, line 13: Naming suggestion:

```suggestion
// failedMachineReclaimThreshold is the duration after which the machine reclaim is assumed to have failed.
const failedMachineReclaimThreshold = time.Minute * 5
```

Gerrit91 (changes requested):

cmd/metal-api/internal/metal/provisioning.go
```diff
@@ -41,9 +47,10 @@ var (
 		ProvisioningEventInstalling:       true,
 		ProvisioningEventBootingNewKernel: true,
 		ProvisioningEventPhonedHome:       true,
+		ProvisioningEventMachineReclaim:   true,
 	}
 	// ProvisioningEventsInspectionLimit The length of how many provisioning events are being inspected for calculating incomplete cycles
-	ProvisioningEventsInspectionLimit = 2 * len(expectedBootSequence) // only saved events count
+	ProvisioningEventsInspectionLimit = 100 // only saved events count
```
Gerrit91, line 53: Can be removed along with the old state machine in this package.

cmd/metal-api/internal/metal/provisioning.go
```diff
@@ -204,9 +212,22 @@ func ProvisioningEventForMachine(log *zap.SugaredLogger, ec *ProvisioningEventCo
...
 		ec.Events = append([]ProvisioningEvent{ev}, ec.Events...)
-		ec.IncompleteProvisioningCycles = ec.CalculateIncompleteCycles(log)
 		ec.Liveliness = MachineLivelinessAlive
 	}
 	ec.TrimEvents(ProvisioningEventsInspectionLimit)
 	return ec
 }
+
+func (c *ProvisioningEventContainer) Validate() error {
+	lastEventTime := c.Events[0].Time
```
Gerrit91, line 222: Can cause index out of bounds.

cmd/metal-api/internal/metal/provisioning.go
```diff
@@ -204,9 +212,22 @@ func ProvisioningEventForMachine(log *zap.SugaredLogger, ec *ProvisioningEventCo
...
 	} else {
 		ec.Events = append([]ProvisioningEvent{ev}, ec.Events...)
-		ec.IncompleteProvisioningCycles = ec.CalculateIncompleteCycles(log)
 		ec.Liveliness = MachineLivelinessAlive
 	}
 	ec.TrimEvents(ProvisioningEventsInspectionLimit)
 	return ec
 }
+
+func (c *ProvisioningEventContainer) Validate() error {
```
Gerrit91, line 221: Unit test would be great.

cmd/metal-api/internal/metal/provisioning.go
```diff
@@ -204,9 +212,22 @@ func ProvisioningEventForMachine(log *zap.SugaredLogger, ec *ProvisioningEventCo
...
+	lastEventTime := c.Events[0].Time
+
+	for _, e := range c.Events {
+		if e.Time.Before(lastEventTime) {
+			return fmt.Errorf("provisioning event container for machine %s is not chronologically sorted", c.ID)
+		}
+
+		lastEventTime = e.Time
+	}
+
```
Gerrit91, line 231: Add check that latest event equals last event time.

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -1855,8 +1854,12 @@ func (r *machineResource) freeMachine(request *restful.Request, response *restfu
 
 	r.send(request, response, http.StatusOK, resp)
 
-	event := string(metal.ProvisioningEventPlannedReboot)
-	_, err = r.ds.ProvisioningEventForMachine(logger, id, event, "freeMachine")
+	ev := metal.ProvisioningEvent{
+		Time:    time.Now(),
+		Event:   metal.ProvisioningEventMachineReclaim,
+		Message: "freeMachine",
```
Gerrit91, line 1860:
```suggestion
		Message: "free machine called",
```

cmd/metal-api/internal/service/v1/machine.go
```diff
@@ -103,6 +103,8 @@ type MachineRecentProvisioningEvents struct {
 	Events                       []MachineProvisioningEvent `json:"log" description:"the log of recent machine provisioning events"`
 	LastEventTime                *time.Time                 `json:"last_event_time" description:"the time where the last event was received" optional:"true"`
 	IncompleteProvisioningCycles string                     `json:"incomplete_provisioning_cycles" description:"the amount of incomplete provisioning cycles in the event container"`
```
Gerrit91, line 105: I will check if there is a field annotation available that marks this deprecated. If it's not possible, just put into the description.

iljarotar:
fsm

cmd/metal-api/internal/fsm/fsm_test.go
```diff
@@ -0,0 +1,315 @@
...
+				LastEventTime:        &now,
+				Events: metal.ProvisioningEvents{
+					{
+						Time:  lastEventTime,
+						Event: metal.ProvisioningEventBootingNewKernel,
+					},
+					{
+						Time:  now,
+						Event: metal.ProvisioningEventPhonedHome,
+					},
```
Gerrit91, line 146: I think in the past the events in the event container were sorted anti-chronologically (latest to oldest). Would it make sense to also call your new `Validate` function in the test to ensure this order is correct?

cmd/metal-api/internal/fsm/fsm_test.go
```diff
@@ -0,0 +1,315 @@
...
+
+func TestHandleProvisioningEvent(t *testing.T) {
+	now := time.Now()
+	lastEventTime := now.Add(-time.Minute * 4)
+	exceedThresholdTime := now.Add(-time.Minute * 10)
+	tests := []struct {
+		event     *metal.ProvisioningEvent
+		container *metal.ProvisioningEventContainer
+		name      string
+		wantErr   error
```
Gerrit91, line 20: There are no tests that have an error. Are there any interesting failure scenarios to cover?

cmd/metal-api/internal/fsm/fsm_test.go
```diff
@@ -0,0 +1,315 @@
...
+				Liveliness:           metal.MachineLivelinessAlive,
+				LastEventTime:        &now,
+				Events: metal.ProvisioningEvents{
+					{
+						Time:  exceedThresholdTime,
+						Event: metal.ProvisioningEventMachineReclaim,
+					},
+				},
+			},
+		},
```
Gerrit91, line 300: Tests for resetting the failed states would be great, too.

cmd/metal-api/internal/fsm/fsm_test.go
Gerrit91: This is from a current DB:

Screenshot from 2022-07-26 08-28-03

Gerrit91 (approved):

Gerrit91:
Found another part in the code where incomplete cycles field is getting used: `machine-service.go:2543`. Should also be replaced with Crashloop and FailedMachineReclaim fields.

Gerrit91:
Maybe we can also implement #146 in this PR. Should be just three lines of code.

iljarotar:
Closes #146

Gerrit91 (approved):

PR #269, opened 2022-03-25
Add events endpoint to be able to batch them

majst01:
Actually we create 2 phone home events per machine from metal-core, one per leaf, which creates a lot of api calls in the metal-api which raises cpu usage. Additionally the ingress treats this sometimes as a DDOS attack and sent GOAWAY back to the caller if a certain amount of request from the same client per time are fired.

### DIFF
--- a/cmd/metal-api/internal/service/machine-service.go
+++ b/cmd/metal-api/internal/service/machine-service.go
@@ -21,6 +21,7 @@ import (
 	"github.com/metal-stack/metal-lib/httperrors"
 	"github.com/metal-stack/metal-lib/pkg/tag"
 	"github.com/metal-stack/metal-lib/zapup"
+	"go.uber.org/multierr"
 	"go.uber.org/zap"
 
 	mdmv1 "github.com/metal-stack/masterdata-api/api/v1"
@@ -328,14 +329,14 @@ func (r machineResource) webService() *restful.WebService {
 		Returns(http.StatusOK, "OK", v1.MachineRecentProvisioningEvents{}).
 		DefaultReturns("Error", httperrors.HTTPErrorResponse{}))
 
-	ws.Route(ws.POST("/events").
+	ws.Route(ws.POST("/event").
 		To(editor(r.addProvisioningEvents)).
 		Operation("addProvisioningEvents").
 		Doc("adds machine provisioning events").
 		Metadata(restfulspec.KeyOpenAPITags, tags).
 		Reads(v1.MachineProvisioningEvents{}).
-		Writes(v1.MachineRecentProvisioningEventsMap{}).
-		Returns(http.StatusOK, "OK", v1.MachineRecentProvisioningEventsMap{}).
+		Writes(v1.MachineRecentProvisioningEventsResponse{}).
+		Returns(http.StatusOK, "OK", v1.MachineRecentProvisioningEventsResponse{}).
 		DefaultReturns("Error", httperrors.HTTPErrorResponse{}))
 
 	ws.Route(ws.POST("/{id}/power/on").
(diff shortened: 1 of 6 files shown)


mwennrich (approved):

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -328,6 +328,16 @@ func (r machineResource) webService() *restful.WebService {
 		Returns(http.StatusOK, "OK", v1.MachineRecentProvisioningEvents{}).
 		DefaultReturns("Error", httperrors.HTTPErrorResponse{}))
 
+	ws.Route(ws.POST("/events").
```
Gerrit91, line 331: Up to this point we have no plural in our API namings, I think it's more consistent to use `/event`.

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -328,6 +328,16 @@ func (r machineResource) webService() *restful.WebService {
...
+		To(editor(r.addProvisioningEvents)).
+		Operation("addProvisioningEvents").
+		Doc("adds machine provisioning events").
+		Metadata(restfulspec.KeyOpenAPITags, tags).
+		Reads(v1.MachineProvisioningEvents{}).
+		Writes(v1.MachineRecentProvisioningEventsMap{}).
+		Returns(http.StatusOK, "OK", v1.MachineRecentProvisioningEventsMap{}).
```
Gerrit91, line 338: If there are many, is it really worth it to send them back?

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -2046,52 +2056,77 @@ func (r machineResource) getProvisioningEventContainer(request *restful.Request,
...
+	if checkError(request, response, utils.CurrentFuncName(), err) {
+		return
+	}
+
+	result := v1.MachineRecentProvisioningEventsMap{}
+	for machineID, event := range requestPayload {
+		ec, err := r.addProvisionEventForMachine(machineID, event)
+		if err != nil {
+			zapup.MustRootLogger().Error("unable to add provisioning event for machine", zap.String("machine", machineID), zap.Error(err))
+			continue
```
Gerrit91, line 2071: Maybe the consumer should get an error back when machine events were not added?

cmd/metal-api/internal/service/machine-service_test.go
```diff
@@ -663,6 +663,45 @@ func TestAddProvisioningEvent(t *testing.T) {
...
+	resp := w.Result()
+	defer resp.Body.Close()
+	require.Equal(t, http.StatusOK, resp.StatusCode, w.Body.String())
+	var result v1.MachineRecentProvisioningEventsMap
+	err = json.NewDecoder(resp.Body).Decode(&result)
+
+	require.NoError(t, err)
+	require.Equal(t, "0", result[machineID].IncompleteProvisioningCycles)
+	require.Len(t, result[machineID].Events, 1)
+	if len(result[machineID].Events) > 0 {
```
Gerrit91, line 699: This is ensured by the `require` in the line before.

cmd/metal-api/internal/service/machine-service.go
majst01: Done

cmd/metal-api/internal/service/machine-service.go
majst01: Changed to a proper response

cmd/metal-api/internal/service/machine-service.go
majst01: still continue to preserve working events, but return if at least one error occurs now

cmd/metal-api/internal/service/machine-service_test.go
majst01: thanks, done

PR #274, opened 2022-04-20
IP free and child network free should be DELETE methods.

Gerrit91:
This fixes and actual bug with newer versions of go-openapi, which leaves out the accept header on empty payload: https://github.com/go-openapi/runtime/issues/231.

We could also make the existing endpoints consume an empty body or `*/*`, but I think this is more consistent to the rest of the API.

### DIFF
--- a/cmd/metal-api/internal/service/ip-service.go
+++ b/cmd/metal-api/internal/service/ip-service.go
@@ -94,6 +94,7 @@ func (ir ipResource) webService() *restful.WebService {
 		Operation("freeIPDeprecated").
 		Doc("frees an ip").
 		Param(ws.PathParameter("id", "identifier of the ip").DataType("string")).
+		Consumes(restful.MIME_JSON, "*/*").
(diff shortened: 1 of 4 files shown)


cmd/metal-api/internal/service/ip-service.go
```diff
@@ -90,6 +90,17 @@ func (ir ipResource) webService() *restful.WebService {
 		DefaultReturns("Error", httperrors.HTTPErrorResponse{}))
 
 	ws.Route(ws.POST("/free/{id}").
+		To(editor(ir.freeIP)).
+		Operation("freeIPDeprecated").
+		Doc("frees an ip").
```
majst01, line 95: `Consume("").` missing ?

Gerrit91: Right, probably makes it even safer to add this, too.

majst01 (approved):

PR #275, opened 2022-04-29
Expose event endpoint for events via grpc

majst01:
Expose event endpoint via grpc. This can be used by metal-hammer to send them directly without the need to go trough metal-core. It also exposes the events endpoint suitable for metal-core to send a batch of events. Both events related send rest endpoints can be removed after metal-hammer and metal-core are updated to use this grpc endpoint

This works so far.


cmd/metal-api/internal/datastore/event.go
```diff
@@ -35,3 +38,41 @@ func (rs *RethinkStore) CreateProvisioningEventContainer(ec *metal.ProvisioningE
 func (rs *RethinkStore) UpsertProvisioningEventContainer(ec *metal.ProvisioningEventContainer) error {
 	return rs.upsertEntity(rs.eventTable(), ec)
 }
+func (rs *RethinkStore) ProvisioningEventForMachine(machineID, event, message string) (*metal.ProvisioningEventContainer, error) {
```
Gerrit91, line 41: I think that it would be better if we factor this code out into a separate event package in order to make it accessible from grpc and rest. This will also give the future-planned FSM a place to live + it keeps the datastore dumb, which we tried to achieve everywhere else.

cmd/metal-api/internal/grpc/grpc-server.go
```diff
@@ -46,6 +54,7 @@ type ServerConfig struct {
 type Server struct {
 	*WaitService
 	*SupwdService
+	*EventService
```
Gerrit91, line 57: For future we should split up server and service code to make this easier to understand and to decouple for more services to come.

cmd/metal-api/internal/datastore/event.go
```diff
@@ -35,3 +38,41 @@ func (rs *RethinkStore) CreateProvisioningEventContainer(ec *metal.ProvisioningE
...
+	now := time.Now()
+	ec.LastEventTime = &now
+
+	ev := metal.ProvisioningEvent{
+		Time:    now,
+		Event:   metal.ProvisioningEventType(event),
+		Message: message,
+	}
+	if ev.Event == metal.ProvisioningEventAlive {
+		zapup.MustRootLogger().Sugar().Debugw("received provisioning alive event", "id", ec.ID)
```
Gerrit91, line 64: When we already touch this, can we pass a logger, too? Would be nice to reduce the usage of zapup.

majst01: done

cmd/metal-api/internal/datastore/event.go
majst01: unsure, only event related logic here

Gerrit91 (approved):

PR #278, opened 2022-05-04
Introduce machine update endpoint.

### LINKED ISSUE #277: Implement machine update
We should have an endpoint to allow adding or removing machine labels. Maybe other fields do make sense as well.

Gerrit91:
Closes #277.

### DIFF
--- a/cmd/metal-api/internal/service/machine-service.go
+++ b/cmd/metal-api/internal/service/machine-service.go
@@ -531,9 +531,8 @@ func (r machineResource) updateMachine(request *restful.Request, response *restf
 	if requestPayload.Description != nil {
 		newMachine.Description = *requestPayload.Description
 	}
-	if requestPayload.Tags != nil {
-		newMachine.Tags = requestPayload.Tags
-	}
+
+	newMachine.Tags = makeMachineTags(&newMachine, requestPayload.Tags)
 
 	err = r.ds.UpdateMachine(oldMachine, &newMachine)
 	if checkError(request, response, utils.CurrentFuncName(), err) {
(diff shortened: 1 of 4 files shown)


cmd/metal-api/internal/service/machine-service.go
```diff
@@ -494,6 +505,52 @@ func (r machineResource) findMachine(request *restful.Request, response *restful
...
+	if oldMachine.Allocation == nil {
+		if checkError(request, response, utils.CurrentFuncName(), fmt.Errorf("only allocated machines can be updated")) {
+			return
+		}
+	}
+
+	newMachine := *oldMachine
+
+	if requestPayload.Name != nil {
+		newMachine.Name = *requestPayload.Name
```
majst01, line 529: unsure if update name is a good idea

Gerrit91: Yes, I also think it can be dangerous in combination with Gardener. On the other hand when you use metalctl in combination with Gardener managed machines a user should know that this will always have dangerous implications.

Gerrit91: From pure perspective of the metal-api and metal machine provisioning I don't think it's dangerous, or?

majst01: I would not enable users to do unnecessary and dangerous operations

Gerrit91: Removed. Editing tags is dangerous though, too! People could remove the Gardener cluster tag, which is important for the MCM.

majst01 (approved):

Issue #209, opened 2021-08-06
Machine allocation succeeded flag never gets set


Gerrit91:
Maybe I miss something, but during investigation of #207 I found that the field `machine.Allocation.Succeeded` field never gets set. I am very confused about this as I thought that this flag toggles in `finalize-allocation`. It doesn't. And also it seems like in the past it did not.

This is actually dangerous as the metal-core uses this flag do decide whether it lets a machine boot the hammer or not: https://github.com/metal-stack/metal-core/blob/master/internal/endpoint/boot.go#L38

Also the GEPM uses this flag in the firewall reconciliation and it's kind of a miracle that the code just works.

vknabel:
Closed due to age. If still needed please reopen.

PR #279, opened 2022-05-06
Grpc boot service

majst01:
Expose Register and Report (former finalizeAllocation) via grpc for direct communication from metal-hammer, therefor all metal-hammer related calls are exposed via grpc boot-service, metal-core lost this responsibility of forwarding them.

Also in this PR is the removal of MachineHardwareExtended and MachineNicExtended and move the neighbors field into the MachineNic struct.

requires: https://github.com/metal-stack/metal-go/pull/83

```BREAKING_CHANGE
Due to breaking changes in the gRPC API, make sure you update to the latest metal-stack minor release before upgrading to this version. If you don't, rebooting all waiting machines by hand is required.
```


cmd/metal-api/internal/grpc/boot-service.go
```diff
@@ -0,0 +1,411 @@
...
+
+	resp := &v1.BootServiceBootResponse{
+		Kernel:       p.BootConfiguration.KernelURL,
+		InitRamDisks: []string{p.BootConfiguration.ImageURL},
+		Cmdline:      &p.BootConfiguration.CommandLine,
+	}
+	b.log.Infow("boot", "resp", resp)
+	return resp, nil
+
+	// if allocateion.Succeed==false, the machine was already in the installation phase but crashed before finalizing allocation
```
majst01, line 79: Needs to be discussed, code was taken from metal-core impl.

majst01:
machine rebooted from outside after `report`

pkg/api/v1/boot.proto
```diff
@@ -0,0 +1,131 @@
+syntax = "proto3";
+
+package api.v1;
+
+option go_package = "./api/v1";
+
+service BootService {
+    // Dhcp is the first dhcp request (option 97). A ProvisioningEventPXEBooting is fired
+    rpc Dhcp(BootServiceDhcpRequest) returns (BootServiceDhcpResponse) {}
+    // Boot is called from pixie once the nachine got the first dhcp response and ipxie asks for subsequent kernel and initrd
```
Gerrit91, line 10:
```suggestion
    // Boot is called from pixie once the machine got the first dhcp response and ipxie asks for subsequent kernel and initrd
```

pkg/api/v1/boot.proto
```diff
@@ -0,0 +1,131 @@
...
+
+message BootServiceBootResponse {
+    string  kernel = 1;
+    repeated string init_ram_disks = 2;
+    optional string cmdline = 3;
+}
+
+message BootServiceRegisterRequest {
+    string uuid = 1;
+	MachineHardware hardware = 2;
```
Gerrit91, line 40: Indention

pkg/api/v1/boot.proto
```diff
@@ -0,0 +1,131 @@
+syntax = "proto3";
+
+package api.v1;
+
+option go_package = "./api/v1";
+
+service BootService {
+    // Dhcp is the first dhcp request (option 97). A ProvisioningEventPXEBooting is fired
+    rpc Dhcp(BootServiceDhcpRequest) returns (BootServiceDhcpResponse) {}
```
Gerrit91, line 9: Maybe it makes sense to directly to call the event endpoint?

cmd/metal-api/internal/grpc/boot-service.go
Gerrit91: I believe that we wanted to be very safe that a machine when the metal-hammer gets booted and the machine still belongs to a user that we do not wipe the disks. But somehow it does not make sense and it also never worked (see #209). I think we can remove it.

cmd/metal-api/internal/grpc/boot-service.go
```diff
@@ -0,0 +1,411 @@
...
+			ipmi.Fru = fru
+		}
+
+	}
+
+	var registerState v1.RegisterState
+	if m == nil {
+		// machine is not in the database, create it
+		name := fmt.Sprintf("%d-core/%s", machineHardware.CPUCores, humanize.Bytes(machineHardware.Memory))
+		descr := fmt.Sprintf("a machine with %d core(s) and %s of RAM", machineHardware.CPUCores, humanize.Bytes(machineHardware.Memory))
```
Gerrit91, line 189: Never seen this somewhere because machine skeleton is created already on PXE boot event. Can be removed.

cmd/metal-api/internal/service/v1/machine.go
```diff
@@ -389,9 +378,16 @@ func NewMachineResponse(m *metal.Machine, s *metal.Size, p *metal.Partition, i *
 
 	nics := MachineNics{}
 	for i := range m.Hardware.Nics {
+		n := m.Hardware.Nics[i]
+		neighs := MachineNics{}
+		for j := range n.Neighbors {
+			neigh := n.Neighbors[j]
+			neighs = append(neighs, MachineNic{MacAddress: string(neigh.MacAddress), Name: neigh.Name})
+		}
```
Gerrit91, line 386: How did we do that before with the extended struct?

pkg/api/v1/boot.proto
majst01: formatted with buf

cmd/metal-api/internal/grpc/boot-service.go
majst01: removed commented code

pkg/api/v1/boot.proto
majst01: No, i wanted to be explicit about whats going on

cmd/metal-api/internal/grpc/boot-service.go
majst01: deleted

cmd/metal-api/internal/service/v1/machine.go
majst01: This was not done before because, this was a BUG?

majst01:
All review findings addressed so far.

majst01:
ensure `metalctl machine rm` still do both, set to PXE and reboot the machine

Makefile
```diff
@@ -5,7 +5,7 @@ MINI_LAB_KUBECONFIG := $(shell pwd)/../mini-lab/.kubeconfig
 
 include $(COMMONDIR)/Makefile.inc
 
-release:: protoc spec check-diff all ;
```
Gerrit91, line 8: Without the protoc target the `check-diff` target will never find changes.

majst01: This is done in CI
https://github.com/metal-stack/metal-api/pull/279/files#diff-023e805b335fe6fb703da891ae5add592e98632ab791a1bbcf2c67a34ecfdc67R27

Gerrit91 (approved):

PR #280, opened 2022-05-09
Ensure image and kernel in Partition are downloadable

majst01:
Found during testing that we dont check if kernel or metal-hammer image are downloadable as we do for machine images.

### DIFF
--- a/cmd/metal-api/internal/service/partition-service_test.go
+++ b/cmd/metal-api/internal/service/partition-service_test.go
@@ -141,6 +141,7 @@ func TestCreatePartition(t *testing.T) {
 	}
 	service := NewPartition(ds, topicCreater)
 	container := restful.NewContainer().Add(service)
+	downloadableFile := "https://upload.wikimedia.org/wikipedia/commons/e/e6/1kb.png"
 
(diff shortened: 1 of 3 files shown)


mwennrich (approved):

cmd/metal-api/internal/service/partition-service_test.go
```diff
@@ -182,7 +187,7 @@ func TestUpdatePartition(t *testing.T) {
 	container := restful.NewContainer().Add(service)
 
 	mgmtService := "mgmt"
-	imageURL := "http://somewhere/image1.zip"
+	downloadableFile := "https://upload.wikimedia.org/wikipedia/commons/e/e6/1kb.png"
```
Gerrit91, line 190: Somehow I don't like so much relying on the internet for unit tests. Sometimes I develop code on a travel or something and then your tests start failing when you're in an airplane. 😅
Don't you think it would be better to direct this to a local endpoint?

majst01: Right, fixed.

Gerrit91 (approved):

Issue #298, opened 2022-06-22
Add connection to headscale


GrigoriyMikhalkin:
`metal-api` should be able to connect to `headscale` server. For that it should be able to accept `headscale-addr` flag.

cc @majst01 @Gerrit91
In case `headscale` address  is provided `metal-api` will be adding VPN for all projects. But what if `headscale` server will go down for some reason? IMO, it should fail to create new firewalls/machine in that case with explicit error that `headscale` server is down. What do you think?

majst01:
Yes if headscale server is configured but down, machine/firewall creation must fail

Gerrit91:
Yep, please make it conditional --> if headscale server is configured it fails, if it's not configured everything works as before

PR #302, opened 2022-06-28
initiate headscale connection

GrigoriyMikhalkin:
Solves https://github.com/metal-stack/metal-api/issues/298

Also it's dependent on https://github.com/juanfont/headscale/pull/666 (PR to headscale to allow to init server with API Key) or similar solution. 

cc @majst01, @Gerrit91 
The only other option to init API Key that i can think of is to use SSH client during Headscale initialization in metal-api, which will be quite messy.

Also, current connection to Headscale is set as insecure, should we/can we establish TLS connection to it?


cmd/metal-api/internal/headscale/client.go
```diff
@@ -0,0 +1,126 @@
...
+
+	h.ctx, h.cancelFunc = context.WithTimeout(context.Background(), cfg.CLI.Timeout)
+	grpcOptions := []grpc.DialOption{
+		grpc.WithBlock(),
+		grpc.WithPerRPCCredentials(tokenAuth{
+			token: apiKey,
+		}),
+	}
+
+	// TODO: set TLS?
```
majst01, line 67: Is not strictly required in my opinion, we connect only inside the same cluster

majst01 (commented):
Dont know why linter fails but this is not important right now.

overall fine so far

PR #304, opened 2022-06-29
Get rid off zapup (part 1) and a bit of test refactoring.

Gerrit91:


cmd/metal-api/internal/service/async-actor.go
```diff
@@ -178,12 +178,12 @@ func (a *asyncActor) disassociateIP(ip *metal.IP, machine *metal.Machine) error
 // to implement more validations: when a machine is deleted the caller has to check if the IP
 // is static or not. If only an IP is freed, the caller has to check if the IP has machine scope.
 func (a *asyncActor) releaseIP(ip metal.IP) error {
-	a.Info("release IP", zap.Any("ip", ip))
+	a.log.Infow("release IP", "ip", ip)
```
majst01, line 181: we must see what is logged then

majst01 (approved):

PR #309, opened 2022-07-09
Use my generic fsm branch

majst01:
I finished the generic PR of looplab/fsm and want to show how this would impact this PR. We should discuss together it this is a gain.

We now also see that there is no differentiation between Events and States, this should be be done as well in metal-api.

@Gerrit91


Gerrit91:
Please open a new pull request against master as soon as there is the v2 looplab version.

PR #312, opened 2022-07-28
Use ipam grpc service instead of library

majst01:
With this PR we move from library usage of go-ipam to a dedicated grpc go-ipam service. This removes the direct dependency to the ipam database as well and makes the overall architecture clearer.

TODO:

- [x] https://github.com/metal-stack/go-ipam/pull/86
- [ ] https://github.com/metal-stack/helm-charts/pull/40
- [ ] https://github.com/metal-stack/metal-roles/pull/99
- [ ] adopt mini-lab
- [ ] update deployment diagram in docs


majst01:
closing because #522  was merged

Issue #311, opened 2022-07-26
Add /v1/vpn/authkey endpoint


GrigoriyMikhalkin:
Add `/v1/vpn/authkey GET` endpoint to the metal-api. This endpoint should only work when VPN is activated, otherwise always return error. It should accept `projectID` as a parameter, check if namespace for project namespace exists and then create new auth key and return it to the user. If namespace for project doesn't exists(means that no firewall is connected to the VPN) -- return error.

This endpoint will be used by `metalctl` to set up VPN connection.

PR #313, opened 2022-07-29
MEP-9: Implement VPN keys utilizing headscale

GrigoriyMikhalkin:
Solves https://github.com/metal-stack/metal-api/issues/311


go.mod
```diff
@@ -14,8 +19,9 @@ require (
 	github.com/google/uuid v1.3.0
 	github.com/grpc-ecosystem/go-grpc-middleware v1.3.0
 	github.com/grpc-ecosystem/go-grpc-prometheus v1.2.0
+	github.com/juanfont/headscale v0.16.0-beta5
 	github.com/looplab/fsm v0.3.0
-	github.com/metal-stack/go-ipam v1.8.5
+	github.com/metal-stack/go-ipam v1.8.7-0.20220620122231-10374d8beeff
```
majst01, line 24: I will check for a proper solution here

cmd/metal-api/internal/headscale/auth.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+// Implements google.golang.org/grpc/credentials.PerRPCCredentials interface
+type tokenAuth struct {
+	token string
+}
+
+func (t tokenAuth) GetRequestMetadata(
+	ctx context.Context,
+	in ...string,
```
Gerrit91, line 14: As params are not used, you can use `_` for their names.

cmd/metal-api/internal/headscale/client.go
```diff
@@ -0,0 +1,238 @@
...
+
+var (
+	apiKeyExpiration = 24 * 90 * time.Hour
+)
+
+type HeadscaleClient struct {
+	client headscalev1.HeadscaleServiceClient
+
+	address             string
+	ControlPlaneAddress string
```
Gerrit91, line 28: Maybe a getter method would be more precise to prevent modification from outside of this package.

cmd/metal-api/internal/headscale/client.go
```diff
@@ -0,0 +1,238 @@
...
+		return fmt.Errorf("failed to connect to headscale server %s: %w", h.address, err)
+	}
+
+	h.client = headscalev1.NewHeadscaleServiceClient(h.conn)
+
+	return
+}
+
+// Refresh API key when apiKeyExpiration reaches it's half life
+func (h *HeadscaleClient) rotateApiKey() {
```
Gerrit91, line 105: Check if we really need the key rotation as discussed.

cmd/metal-api/internal/headscale/client.go
```diff
@@ -0,0 +1,238 @@
...
+				break
+			}
+		}
+	}
+}
+
+// Expires current API Key and replaces it with a new one
+func (h *HeadscaleClient) replaceApiKey() (err error) {
+	h.mu.Lock()
+	defer h.mu.Unlock()
```
Gerrit91, line 135: Probably it's fine to remove these locks from the functions, it's really unlikely that this will cause problems.

cmd/metal-api/internal/service/firewall-service.go
```diff
@@ -215,6 +231,42 @@ func (r *firewallResource) allocateFirewall(request *restful.Request, response *
 	r.send(request, response, http.StatusOK, resp)
 }
 
+func (r firewallResource) setVPNConfigInSpec(allocationSpec *machineAllocationSpec) error {
+	if r.headscaleClient == nil {
+		return nil
+	}
+
+	ctx := context.Background()
```
Gerrit91, line 239: Context from the restful request should be used.

cmd/metal-api/internal/service/firewall-service.go
```diff
@@ -215,6 +231,42 @@ func (r *firewallResource) allocateFirewall(request *restful.Request, response *
...
+
+	// Get project VPN namespace
+	projectID := allocationSpec.ProjectID
+	p, err := r.mdc.Project().Get(ctx, &mdmv1.ProjectGetRequest{Id: projectID})
+	if err != nil {
+		return err
+	}
+	if p.GetProject() == nil {
+		return fmt.Errorf("Project with ID %s wasn't found", projectID)
```
Gerrit91, line 248:
```suggestion
		return fmt.Errorf("project with ID %s wasn't found", projectID)
```

Gerrit91: In general we use only small letters in error messages and log lines.

cmd/metal-api/internal/service/firewall-service.go
```diff
@@ -215,6 +231,42 @@ func (r *firewallResource) allocateFirewall(request *restful.Request, response *
...
+	}
+
+	// Try to create namespace in Headscale DB
+	if err = r.headscaleClient.CreateNamespace(p.Project.Name); err != nil {
```
Gerrit91, line 252: Decided to use the project ID for the namespace name, this way you also do not need to fetch the project from the masterdata api.

cmd/metal-api/internal/service/firewall-service.go
```diff
@@ -215,6 +231,42 @@ func (r *firewallResource) allocateFirewall(request *restful.Request, response *
...
+		return fmt.Errorf("failed to create new VPN namespace for the Project: %w", err)
+	}
+
+	expiration := time.Now().Add(90 * 24 * time.Hour)
```
Gerrit91, line 256: If we do not need the rotation and it's really just a join token that's being requested from the headscale server, we can reduce the expiration time to two hours or something (something small).

cmd/metal-api/internal/service/project-service.go
```diff
@@ -25,7 +25,11 @@ type projectResource struct {
 }
 
 // NewProject returns a webservice for project specific endpoints.
-func NewProject(log *zap.SugaredLogger, ds *datastore.RethinkStore, mdc mdm.Client) *restful.WebService {
+func NewProject(
```
Gerrit91, line 28: Nothing really changed, also we do not use C-style function declaration style so much... :D

cmd/metal-api/internal/service/v1/vpn.go
```diff
@@ -0,0 +1,6 @@
+package v1
+
+type VPNResponse struct {
+	Address string `json:"address" description:"Address of VPN's control plane"`
+	AuthKey string `json:"auth_key" description:"Auth key to connect to the VPN"`
```
Gerrit91, line 5: Capitalization maybe.

cmd/metal-api/internal/service/vpn-service.go
```diff
@@ -0,0 +1,108 @@
...
+		Consumes(restful.MIME_JSON).
+		Produces(restful.MIME_JSON)
+
+	tags := []string{"vpn"}
+
+	ws.Route(ws.GET("/authkey/{pid}").
+		To(admin(r.getVPNAuthKey)).
+		Operation("getVPNAuthKey").
+		Doc("create auth key to connect to Project's VPN").
+		Param(ws.PathParameter("pid", "identifier of the Project").DataType("string")).
```
Gerrit91, line 58:
```suggestion
		Param(ws.PathParameter("pid", "identifier of the project").DataType("string")).
```

cmd/metal-api/internal/service/vpn-service.go
```diff
@@ -0,0 +1,108 @@
...
+		DefaultReturns("Error", httperrors.HTTPErrorResponse{}))
+
+	return ws
+}
+
+func (r *vpnResource) getVPNAuthKey(request *restful.Request, response *restful.Response) {
+	ctx := context.Background()
+	pid := request.PathParameter("pid")
+
+	p, err := r.mdc.Project().Get(ctx, &mdmv1.ProjectGetRequest{Id: pid})
```
Gerrit91, line 71: Not necessary when using project IDs as identifier.

cmd/metal-api/internal/service/vpn-service.go
```diff
@@ -0,0 +1,108 @@
...
+
+	if ok := r.headscaleClient.NamespaceExists(p.Project.Name); !ok {
+		r.sendError(
+			request, response,
+			httperrors.NotFound(fmt.Errorf("VPN namespace doesn't exist for Project with ID %s", pid)),
+		)
+		return
+	}
+
+	expiration := time.Now().Add(90 * 24 * time.Hour)
```
Gerrit91, line 92: Can be much smaller.

cmd/metal-api/internal/service/vpn-service.go
```diff
@@ -0,0 +1,108 @@
...
+func (r *vpnResource) webService() *restful.WebService {
+	ws := new(restful.WebService)
+	ws.
+		Path(BasePath + "v1/vpn").
+		Consumes(restful.MIME_JSON).
+		Produces(restful.MIME_JSON)
+
+	tags := []string{"vpn"}
+
+	ws.Route(ws.GET("/authkey/{pid}").
```
Gerrit91, line 54: `POST` would be a bit better because we will audit these requests and then you could also let the user pass an expiration date in the request payload.

Dockerfile
```diff
@@ -3,4 +3,4 @@ FROM metalstack/builder:latest as builder
 FROM alpine:3.16
 RUN apk -U add ca-certificates
 COPY --from=builder /work/bin/metal-api /metal-api
-CMD ["/metal-api"]
+CMD ["/metal-api", "--headscale-api-key=t-c59Xz83A.7gfBg50Cfl_YVkBlwHnduD-aJXBKKMaXRXtjvqO_aOo", "--headscale-addr=headscale.headscale:50443", "--headscale-cp-addr=http://headscale.172.17.0.1.nip.io:8080"]
```
Gerrit91, line 6: Needs to be reverted, I think?

cmd/metal-api/internal/service/firewall-service.go
GrigoriyMikhalkin: Unfortunately we need the rotation. Tested connection to headscale after token is expired(initial connection with non-expired token) -- it stops working after that.

cmd/metal-api/internal/headscale/client.go
GrigoriyMikhalkin: The key rotation is a little bit more complex than i realized before. First of all, yes we absolutely need it as expired key doesn't work completely. Another concern is that the initial key will be stored in a `Secret`, so when pod restarted it will always try to use key from this secret even if it's expired.

I think the ideal solution would be to rotate that Secret. So the key rotation task would be outside of the metal-api scope. WDYT?

majst01 (approved):

cmd/metal-api/internal/headscale/client.go
```diff
@@ -0,0 +1,124 @@
...
+
+	return true
+}
+
+func (h *HeadscaleClient) CreateNamespace(name string) error {
+	req := &headscalev1.CreateNamespaceRequest{
+		Name: name,
+	}
+	_, err := h.client.CreateNamespace(h.ctx, req)
+	if err != nil && !errors.Is(headscalecore.Error("Namespace already exists"), err) {
```
Gerrit91, line 100: Update to error const from headscale.

cmd/metal-api/internal/service/firewall-service.go
```diff
@@ -215,6 +229,31 @@ func (r *firewallResource) allocateFirewall(request *restful.Request, response *
...
+		return nil
+	}
+
+	// Try to create namespace in Headscale DB
+	projectID := allocationSpec.ProjectID
+	if err := r.headscaleClient.CreateNamespace(projectID); err != nil {
+		return fmt.Errorf("failed to create new VPN namespace for the project: %w", err)
+	}
+
+	expiration := time.Now().Add(90 * 24 * time.Hour)
```
Gerrit91, line 243:
```suggestion
	expiration := time.Now().Add(2 * time.Hour)
```

cmd/metal-api/internal/service/vpn-service.go
```diff
@@ -0,0 +1,102 @@
...
+
+	mdm "github.com/metal-stack/masterdata-api/pkg/client"
+	"github.com/metal-stack/metal-api/cmd/metal-api/internal/headscale"
+	v1 "github.com/metal-stack/metal-api/cmd/metal-api/internal/service/v1"
+	"github.com/metal-stack/metal-lib/httperrors"
+)
+
+type vpnResource struct {
+	webResource
+	mdc             mdm.Client
```
Gerrit91, line 21: Unused

Gerrit91 (approved):

PR #314, opened 2022-08-08
Factor out states

Gerrit91:


cmd/metal-api/internal/fsm/fsm.go
```diff
@@ -166,125 +24,70 @@ func HandleProvisioningEvent(log *zap.SugaredLogger, ec *metal.ProvisioningEvent
...
-}
+	var (
+		clone       = *ec
+		container   = &clone
+		latestEvent = func() string {
+			if len(container.Events) != 0 {
+				return container.Events[0].Event.String()
+			}
+			return ""
+		}()
```
iljarotar, line 35:
```suggestion
		latestEvent = len(container.Events) != 0 ? container.Events[0].Event.String() : ""
```

What do you think about this syntax?

Gerrit91: Put into separate function.

iljarotar (approved):

PR #317, opened 2022-08-09
Persist user email who locked or reserved a machine in DB #303

### LINKED ISSUE #303: Persist user name who locked or reserved a machine in DB
When someone locks a machine, it would be really helpful to save the name of the person who locked the machine.

vknabel:
When locking or reserving a specific machine, the metal-api additionally tracks the email address of the state change issuer.
This allows other users to directly contact the issuer in case they need additional information. Closes #303.

This PR requires additional changes in other repositories:

- metal-stack/metal-api#317: tracking the issuer and provide access via api
- metal-stack/metal-go#89: reflects the added issuer field of the api
- metal-stack/metalctl _(pr after merge)_: bump metal-go dependency (requires metal-go release)

As a new response field has been added to the api, this probably requires a minor update.

**How to test?**

_To use the updated metalctl, bump the metal-go version in `go.mod`, e.g. with `replace github.com/metal-stack/metal-go => ../metal-go`_

- assumes you have a running metal stack (e.g. mini-lab) and the fitting environment (`eval $(make dev-env)` from mini-lab)
- with new metalctl against old metal-api: `go run . machine describe $id` should work except issuer
- deploy this metal-api (`make mini-lab-push`)
- use the metalctl with a locally bumped metal-go version
- with new or old metalctl: `go run . machine lock $id`
- with new metalctl to see issue: `go run . machine describe $id`


cmd/metal-api/internal/service/machine-service.go
```diff
@@ -628,6 +628,16 @@ func (r *machineResource) setMachineState(request *restful.Request, response *re
...
 
+	user, err := r.userGetter.User(request.Request)
+	if err != nil {
+		r.sendError(request, response, defaultError(err))
+		return
+	}
+	var userId string
+	if user != nil {
+		userId = user.EMail
+	}
```
Gerrit91, line 639: You can assume the user is not nil when there was no `err` returned.

Gerrit91 (commented):
Great PR, nice, clean and with test.

cmd/metal-api/internal/service/machine-service.go
vknabel: Thanks! Removed the check and the corresponding unit test

vknabel:
In the latest commits I declared the `Issue` as optional. Otherwise new versions of metalctl would not be downwards compatible.
Ready for final review :)

Gerrit91:
When you remove the lock, the issuer is still in the database. Maybe we can revert it to empty string, too?

Gerrit91:
Ah. This is done by the client, right? Then, I think we are good to go. :)

vknabel:
Currently the client only tells the metal-api to set the state. The metal-api itself adds the issuer to the executing user. And the issuer is exposed as readonly value.

When unlocking, the machine state _and_ the issuer will be set by the api to the currently unlocking user, which might be useful information (e.g. "Why did you unlock the machine?"). But due to a missing timestamp probably just with limited value.

I could add a special case for `AvailableState` which sets the issuer to `""` if you wish.

Gerrit91:
I think the metal-hammer fiddles around with the machine state, too (which BTW I think is far from ideal but it goes beyond the topic of this PR). I think the `AvailableState` is equal to `""` (empty state). So maybe just wipe the user when `""` gets set?

vknabel:
Correct, the `MetalHammerVersion` is also stored within the state, but tbh, I am not sure what it represents.

Yes, the `AvailableState` is the empty state `""`. With the latest commit, the issuer will now be empty, when the machine is available. And as it is a special case, there is a unit test for that.

Gerrit91 (approved):

Gerrit91:
The `MetalHammerVersion` just shows the version of the metal-hammer at the point in time when the machine is registered, which can be interesting for the operator. The metal-hammer does a reboot after three days in order to update to the version that is configured in the partition configuration.

PR #323, opened 2022-08-31
use bmc tags from metal-lib

majst01:
depends on:

- [ ] https://github.com/metal-stack/metal-lib/pull/57
- [ ] https://github.com/metal-stack/metal-bmc/pull/42


Gerrit91:
This relates to the NSQ integration, which we want to drop with the new metal-apiserver introduced by MEP-4.

PR #329, opened 2022-09-07
Fix switch replacement.

mreiger:


Gerrit91:
If you still want to provide a test, please open a new pull request. This should be in the next release, so I merge it immediately.

Gerrit91 (approved):

PR #336, opened 2022-09-16
check if firewall is connected to VPN

GrigoriyMikhalkin:


cmd/metal-api/internal/headscale/client.go
```diff
@@ -119,6 +119,29 @@ func (h *HeadscaleClient) CreatePreAuthKey(namespace string, expiration time.Tim
 	return resp.PreAuthKey.Key, nil
 }
 
+func (h *HeadscaleClient) DescribeMachine(hostname, projectID string) (connected bool, err error) {
```
majst01, line 122: LGTM, the only downside actually is that hostname is not guaranteed to be unique in a project. It is the case for our clusters but in the pure metal use case it is not. Should be at least documented.

majst01: One idea would be to add a tag from the firewall during connecting with:

`tailscale up --advertise-tags tag:metal-stack.io/machineid=<id of the firewall> `

and then check if the machineid tag with exists

majst01: Yes sure

GrigoriyMikhalkin: Situation is even more complicated then i initially thought. Headscale API refuses to return any tags if ACL policy isn't set. So tags are very tightly connected to ACLs.

cmd/metal-api/internal/headscale/client.go
```diff
@@ -119,6 +119,30 @@ func (h *HeadscaleClient) CreatePreAuthKey(namespace string, expiration time.Tim
...
+	req := &headscalev1.ListMachinesRequest{
+		Namespace: projectID,
+	}
+	resp, err := h.client.ListMachines(h.ctx, req)
+	if err != nil || resp == nil {
+		return false, fmt.Errorf("failed to list machines: %w", err)
+	}
+
+	for _, m := range resp.Machines {
+		if m.Name == machineid {
```
majst01, line 132: i thought a check for the tag containing the machineid is now possible, why is the name now equal to machineid ?

GrigoriyMikhalkin: https://github.com/metal-stack/metal-api/pull/336#discussion_r975745133, unfortunately it isn't possible if ACL policy isn't set.

majst01 (approved):

PR #338, opened 2022-10-18
debug logging in DescribeMachine method

GrigoriyMikhalkin:


majst01:
No required, works as implemented.

PR #341, opened 2022-10-24
do not return an error if firewall is not connected

majst01:


cmd/metal-api/internal/service/firewall-service.go
```diff
@@ -139,8 +139,7 @@ func (r *firewallResource) findFirewall(request *restful.Request, response *rest
 	if fw.Allocation != nil && r.headscaleClient != nil {
 		connected, err = r.headscaleClient.DescribeMachine(fw.ID, fw.Allocation.Project)
 		if err != nil {
-			r.sendError(request, response, defaultError(err))
-			return
+			r.log.Infow("unable to get headscale connected state, ignoring", "firewall", fw.ID)
```
Gerrit91, line 142: In the long term I am not 100% confident if we should dynamically retrieve this information with an individual call from the backend. I think it would be better to just have a `connected` field that is stored in the database, which is updated asynchronously.

Gerrit91: Also I think this should be error level.

Gerrit91 (approved):

PR #351, opened 2022-10-26
WIP: support NIC aliases as alternative to MAC addresses

GrigoriyMikhalkin:
cc @robertvolkmann 

Updated both metal API and GRPC boot API(for metal-hammer).


cmd/metal-api/internal/metal/network.go
```diff
@@ -14,6 +14,7 @@ type MacAddress string
 type Nic struct {
 	MacAddress MacAddress `rethinkdb:"macAddress" json:"macAddress"`
 	Name       string     `rethinkdb:"name" json:"name"`
+	Alias      string     `rethinkdb:"alias" json:"alias"`
```
majst01, line 17: Just a wild idea:

make alias a pointer of string, add to Nic a method `nic.ByIdentifier()` which internally checks for either Mac or Alias.

Then ByMac and ByAlias can be made private

majst01 (changes requested):

cmd/metal-api/internal/metal/network.go
robertvolkmann: Does the metal API really have to know if the identifier is a MAC address or some weird string?

My gut feeling tells me that we should replace `MacAddress` simply by `Identifier`. For backward compatibility, we should introduce a new api version and ship the old and the new one several releases. Similar for the rethinkdb.

GrigoriyMikhalkin: Introducing new `Indentifier` field sounds like a good idea, yep.

Gerrit91:
Before merging, we should manually test this PR:

- [ ] Run in mini-lab
- [ ] Deploy to test environment
  - [ ] Ensure metal-core restart is working (switch registration)
  - [ ] Ensure older metal-core compatible with this PR
- [ ] Delete machine from database and see that it comes back properly

Gerrit91 (approved):

Gerrit91 (approved):

PR #352, opened 2022-10-28
Add hostname field to Boot Register request

GrigoriyMikhalkin:
cc @robertvolkmann 

Added `hostname` field to Boot gRPC service Register request. The logic handles it in such way that the field is expected only from `metal-hammer` at the moment. With possibility of extending functionality of this field in a future.


cmd/metal-api/internal/metal/network.go
```diff
@@ -166,3 +167,13 @@ func (nics Nics) ByName() map[string]*Nic {
 	}
 	return res
 }
+
+func (nics Nics) FilterByHostname(hostname string) (res Nics) {
+	for i, n := range nics {
```
majst01, line 172: For backward compatibility we should return all nics if hostname is empty string

WDYT

Gerrit91: Also looks breaking to me. The switch nics do not have a hostname set at the moment, how is this supposed to find any nics for existing environments?

GrigoriyMikhalkin: @Gerrit91 I think we already store hostnames for switch NICs. The current version of metal-hammer doesn't send the hostname -- but omitting hostname filter for NICs should work as current switches will be identified by MAC.

Gerrit91: Looked it up again. 
- Switches do not have hostnames stored in their nics
- However, you do not look up the hostname in a switch nic but only for machine nic  
- The `ConnectMachine` function is only called during register, so these changes should be fine with an updated metal-hammer version

PR #356, opened 2022-11-08
auto-scaling available machine pool size

iljarotar:


cmd/metal-api/internal/datastore/machine.go
```diff
@@ -429,6 +429,22 @@ func (rs *RethinkStore) ListMachines() (metal.Machines, error) {
 	return ms, err
 }
 
+func (rs *RethinkStore) listMachinesInPartition(partitionid, sizeid string) (metal.Machines, error) {
```
Gerrit91, line 432: I think I would prefer re-using the `SearchMachines` function instead of creating further functions for that. Maybe you need to extend the search query, but this would also be beneficial for other clients.

cmd/metal-api/internal/datastore/event.go
```diff
@@ -62,7 +79,121 @@ func (rs *RethinkStore) ProvisioningEventForMachine(log *zap.SugaredLogger, even
...
+		Type: metal.COMMAND,
+		Cmd: &metal.MachineExecCommand{
+			Command:         cmd,
+			TargetMachineID: m.ID,
+			IPMI:            &m.IPMI,
+		},
+	}
+
+	newMachine := *m
+	newMachine.State = metal.MachineState{Value: state}
```
Gerrit91, line 186: A proper description should be provided in order to see that the scaling algorithm put the machine into this state.

cmd/metal-api/internal/datastore/event.go
```diff
@@ -62,7 +79,121 @@ func (rs *RethinkStore) ProvisioningEventForMachine(log *zap.SugaredLogger, even
...
+	} else {
+		mathrand.Seed(time.Now().UnixNano())
+		// nolint
+		idx = mathrand.Intn(len(machines))
+	}
+
+	return machines[idx]
+}
+
+func (rs *RethinkStore) machineCommand(logger *zap.SugaredLogger, m *metal.Machine, publisher bus.Publisher, cmd metal.MachineCommand, state metal.MState) error {
```
Gerrit91, line 175: Putting this here looks like code duplication from functionality in the service package. We once had the objective to make the datastore layer as dumb as possible and keep specific business functionality in the service layer. Maybe it's better to pass down functions to narrow down code duplication and try to follow this objective if possible.

cmd/metal-api/internal/datastore/event.go
```diff
@@ -62,7 +79,121 @@ func (rs *RethinkStore) ProvisioningEventForMachine(log *zap.SugaredLogger, even
 	}
 
 	newEC.TrimEvents(100)
-
 	err = rs.UpsertProvisioningEventContainer(newEC)
+	if err != nil {
+		return nil, err
+	}
+
+	if config.Message == states.FSMMessageEnterWaitingState || config.Message == states.FSMMessageLeaveWaitingState {
```
Gerrit91, line 87: Can this be moved into the state machine? I think it increases coupling in the code to have such "after FSM conditions". The main reason to put this here is because was probably because the datastore has the `AdjustNumberOfWaitingMachines` functions, but can't this be passed into the FSM as a function or something like that?

cmd/metal-api/internal/datastore/event.go
```diff
@@ -62,7 +79,121 @@ func (rs *RethinkStore) ProvisioningEventForMachine(log *zap.SugaredLogger, even
...
+
+	allMachines, err := rs.listMachinesInPartition(m.PartitionID, m.SizeID)
+	if err != nil {
+		return 0
+	}
+
+	return actual - int(math.Round(float64(len(allMachines))*float64(r)/100))
+}
+
+func randomMachine(machines metal.Machines) metal.Machine {
```
Gerrit91, line 161: This code was taken from the `datastore/machine.go`, right? If yes, then it's code duplication and should be refactored. I am wondering though why we need a random value from `crypto/rand`?

cmd/metal-api/internal/datastore/event.go
```diff
@@ -62,7 +79,121 @@ func (rs *RethinkStore) ProvisioningEventForMachine(log *zap.SugaredLogger, even
...
+	}
+
+	if config.Message == states.FSMMessageEnterWaitingState || config.Message == states.FSMMessageLeaveWaitingState {
+		err = rs.AdjustNumberOfWaitingMachines(log, publisher, machine)
+	}
+
 	return newEC, err
 }
+
+func (rs *RethinkStore) AdjustNumberOfWaitingMachines(log *zap.SugaredLogger, publisher bus.Publisher, machine *metal.Machine) error {
```
Gerrit91, line 94: Unit tests would be great.

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -1820,7 +1820,7 @@ func ResurrectMachines(ds *datastore.RethinkStore, publisher bus.Publisher, ep *
 			continue
 		}
 
-		if provisioningEvents.Liveliness == metal.MachineLivelinessDead && time.Since(*provisioningEvents.LastEventTime) > metal.MachineResurrectAfter {
+		if provisioningEvents.Liveliness == metal.MachineLivelinessDead && time.Since(*provisioningEvents.LastEventTime) > metal.MachineResurrectAfter && m.State.Value != metal.ShutdownState {
```
Gerrit91, line 1823: It's easier to understand moving this condition further up in the loop and guard code with `continue` as done in the lines above.

Gerrit91 (commented):
Great ideas in there! Left comments for discussion.

cmd/metal-api/internal/fsm/events.go
```diff
@@ -159,5 +159,8 @@ func eventCallbacks(config *states.StateConfig) fsm.Callbacks {
 		callbacks["enter_"+name] = state.OnTransition
 	}
 
+	waiting := allStates[states.Waiting.String()].(*states.WaitingState)
```
Gerrit91, line 162: Maybe a little cleaner and more extendable for the future would be to extend the `FSMState` interface with a function `OnLeave(e *fsm.Event)`. Then we can add the callback inside the loop for all states:

```golang
	for name, state := range allStates {
		callbacks["enter_"+name] = state.OnTransition
		callbacks["leave_"+name] = state.OnLeave
	}
```

In order not having to implement the function for all states, we can also consider introducing a small helper struct:

```
type noopState struct{}

func (_ noopState) OnLeave(e *fsm.Event)      {}
func (_ noopState) OnTransition(e *fsm.Event) {}
``` 

Which we can then add to all other states (except Waiting of course):

```
type AliveState struct {
	noopState
	log       *zap.SugaredLogger
	container *metal.ProvisioningEventContainer
	event     *metal.ProvisioningEvent
}
``` 

This way, we can override whenever we want and also add other callbacks in the same way in the future.

cmd/metal-api/internal/fsm/states/waiting.go
```diff
@@ -2,21 +2,28 @@ package states
...
 	}
 }
 
 func (p *WaitingState) OnTransition(e *fsm.Event) {
-	appendEventToContainer(p.event, p.container)
+	appendEventToContainer(p.config.Event, p.config.Container)
+
+	if p.config.Scaler != nil {
+		e.Err = p.config.Scaler.AdjustNumberOfWaitingMachines(p.config.Partition)
+	}
```
Gerrit91, line 22: Can you explain to me why we need to do this `OnTransition` and then again on `OnLeave`? Why not only `OnLeave`?

cmd/metal-api/internal/scaler/poolscaler.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+// AdjustNumberOfWaitingMachines compares the number of waiting machines to the required pool size of the partition and shuts down or powers on machines accordingly
+func (p *PoolScaler) AdjustNumberOfWaitingMachines(partition *metal.Partition) error {
+	waitingMachines, err := p.manager.WaitingMachines()
+	if err != nil {
+		return err
+	}
+
+	poolSizeExcess := 0
```
Gerrit91, line 47: This seems to be an ineffective assignment, it's always overwritten by the next line.

cmd/metal-api/internal/scaler/poolscaler.go
```diff
@@ -0,0 +1,178 @@
...
+	poolSizeExcess = p.calculatePoolSizeExcess(len(waitingMachines), partition.WaitingPoolMinSize, partition.WaitingPoolMaxSize)
+
+	if poolSizeExcess == 0 {
+		return nil
+	}
+
+	if poolSizeExcess > 0 {
+		p.log.Infow("shut down spare machines", "number", poolSizeExcess, "partition", p)
```
Gerrit91, line 55:
```suggestion
		p.log.Infow("shut down spare machines", "number", poolSizeExcess, "partition", partition.ID)
```

cmd/metal-api/internal/scaler/poolscaler.go
```diff
@@ -0,0 +1,178 @@
...
+		}
+	} else {
+		shutdownMachines, err := p.manager.ShutdownMachines()
+		if err != nil {
+			return err
+		}
+
+		poolSizeExcess = int(math.Abs(float64(poolSizeExcess)))
+		if len(shutdownMachines) < poolSizeExcess {
+			p.log.Infow("not enough machines to meet waiting pool size; power on all remaining", "number", len(shutdownMachines), "partition", p)
```
Gerrit91, line 68:
```suggestion
			p.log.Infow("not enough machines to meet waiting pool size; power on all remaining", "number", len(shutdownMachines), "partition", partition.ID)
```

cmd/metal-api/internal/scaler/poolscaler.go
```diff
@@ -0,0 +1,178 @@
...
+		if err != nil {
+			return err
+		}
+	} else {
+		shutdownMachines, err := p.manager.ShutdownMachines()
+		if err != nil {
+			return err
+		}
+
+		poolSizeExcess = int(math.Abs(float64(poolSizeExcess)))
```
Gerrit91, line 66: Is taking the absolute value really required? Isn't the value always >= 0?

cmd/metal-api/internal/scaler/poolscaler.go
```diff
@@ -0,0 +1,178 @@
...
+		if len(shutdownMachines) < poolSizeExcess {
+			p.log.Infow("not enough machines to meet waiting pool size; power on all remaining", "number", len(shutdownMachines), "partition", p)
+			err = p.powerOnMachines(shutdownMachines, len(shutdownMachines))
+			if err != nil {
+				return err
+			}
+			return nil
+		}
+
+		p.log.Infow("power on missing machines", "number", poolSizeExcess, "partition", p)
```
Gerrit91, line 76:
```suggestion
		p.log.Infow("power on missing machines", "number", poolSizeExcess, "partition", partition.ID)
```

cmd/metal-api/internal/scaler/poolscaler.go
```diff
@@ -0,0 +1,178 @@
...
+		ShutdownMachines() (metal.Machines, error)
+
+		Shutdown(m *metal.Machine) error
+		PowerOn(m *metal.Machine) error
+	}
+)
+
+func NewPoolScaler(log *zap.SugaredLogger, manager MachineManager) *PoolScaler {
+	return &PoolScaler{
+		log:     log,
```
Gerrit91, line 35: If the pool scaler gets the partition for the constructor, you can even add this to all logs by initializing the log adding `.With()`.

cmd/metal-api/internal/scaler/poolscaler.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	return nil
+}
+
+// calculatePoolSizeExcess checks if there are less waiting machines than minRequired or more than maxRequired
+// if yes, it returns the difference between the actual amount of waiting machines and the average of minRequired and maxRequired
+// if no, it returns 0
+func (p *PoolScaler) calculatePoolSizeExcess(actual int, minRequired, maxRequired string) int {
+	if strings.Contains(minRequired, "%") != strings.Contains(maxRequired, "%") {
```
Gerrit91, line 90: `strings.HasSuffix` could be a bit more appropriate.

Feels like handling the "percent or int" property makes the function a little harder to read. Maybe it would be a nice idea to factor it out into a small helper type called `ScalerRange` in the `metal/partition.go` with small functions like `IsPercentage` or even just `GetMin` and `GetMax` (figuring out / calculating the value by itself)? You can then also put your validations in there.

cmd/metal-api/internal/scaler/poolscaler.go
```diff
@@ -0,0 +1,178 @@
...
+func randomIndices(n, k int) []int {
+	indices := make([]int, n)
+	for i := range indices {
+		indices[i] = i
+	}
+
+	for i := 0; i < (n - k); i++ {
+		rand.Seed(time.Now().UnixNano())
+		r := rand.Intn(len(indices)) //nolint:gosec
+		indices = append(indices[:r], indices[r+1:]...)
```
Gerrit91, line 150: There is also a `rand.Shuffle`. :) 

```
rand.Shuffle(len(indices), func(i, j int) { a[i], a[j] = a[j], a[i] })
```

cmd/metal-api/internal/scaler/poolscaler_test.go
```diff
@@ -0,0 +1,132 @@
...
+	tests := []struct {
+		name      string
+		m         *metal.Machine
+		partition *metal.Partition
+		mockFn    func(mock *MockMachineManager)
+		wantErr   bool
+	}{
+		{
+			name:      "waiting machines match pool size; do nothing",
+			m:         &metal.Machine{SizeID: ""},
```
Gerrit91, line 20: `SizeID` does not have to be defined to it's anyway default value? Even seems like `m` is not at all required for the test.

cmd/metal-api/internal/datastore/poolsize.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func (m *manager) WaitingMachines() (metal.Machines, error) {
+	stateValue := string(metal.AvailableState)
+	waiting := true
+	q := MachineSearchQuery{
+		PartitionID: &m.partitionid,
+		SizeID:      &m.sizeid,
+		StateValue:  &stateValue,
+		Waiting:     &waiting,
```
Gerrit91, line 39: I am not 100% sure about it, but at first glance it seems problematic to use "waiting" to figure out the amount of waiting machines as there can be many reasons why a machine is not waiting but it's still kind of "on the way of becoming available soon" (the field is toggled when the metal-hammer establishes a gRPC connection to metal-api). 

Potentially this could lead to scenarios if many allocations occur at the same time, the pool scaler starts more and more machines overshooting the threshold as soon as the machines change into waiting state.

I think it's safer to only rely on state values, isn't it?

cmd/metal-api/internal/service/partition-service.go
```diff
@@ -207,6 +209,26 @@ func (r *partitionResource) createPartition(request *restful.Request, response *
 		commandLine = *requestPayload.PartitionBootConfiguration.CommandLine
 	}
 
+	var waitingpoolminsize string
+	if requestPayload.PartitionWaitingPoolMinSize != nil {
+		waitingpoolminsize = *requestPayload.PartitionWaitingPoolMinSize
+	} else {
+		waitingpoolminsize = "0%"
```
Gerrit91, line 216: Allow both to be zero would be cleaner I think because existing database entries are then also "valid" and we do not need a migration. If you consider the `ScalerRange` type as proposed earlier on, you could also add `IsDisabled` and then skip initializing the PoolScaler before passing it to the FSM.

cmd/metal-api/internal/datastore/machine.go
```diff
@@ -58,6 +58,7 @@ type MachineSearchQuery struct {
 
 	// state
 	StateValue *string `json:"state_value" optional:"true"`
+	Waiting    *bool   `json:"waiting" optional:"true"`
```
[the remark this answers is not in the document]
iljarotar: Removed the waiting field according to your comment below:
image

cmd/metal-api/internal/fsm/states/waiting.go
iljarotar: `OnTransition` happens when waiting state is entered, which should also be handled as far as I understand. Maybe we should call this `OnEnter` instead.

cmd/metal-api/internal/scaler/poolscaler.go
iljarotar: The function `calculatePoolSizeExcess` returns `actual` number of waiting machines minus `average` of max and min required waiting machines. It will be the positive value of machines, that need to be shut down, if pool size is exceeded and the negative value of machines, that need to be powered on, if more machines are needed.

cmd/metal-api/internal/fsm/events.go
Gerrit91: Cool, so now we can remove these two lines, right?

cmd/metal-api/internal/metal/partition.go
```diff
@@ -29,3 +36,62 @@ func (sz Partitions) ByID() PartitionMap {
 	}
 	return res
 }
+
+type ScalerRange struct {
+	WaitingPoolMinSize string
+	WaitingPoolMaxSize string
```
Gerrit91, line 42: Could be a little better to continue saving the min and max in the database and decouple the `ScalerRange` from the database. Advantage would be: We can have a constructor that prevents outer modification of the min/max values and the constructor could also do the parsing / validation. This way, we error out as early as possible and from that point we can be sure that everything will be alright with the outputs. 

Don't know if you like it: Changing the signature to `Min(of int)` and `Max(of int)`, the functions can also do the calculation the percentage value if necessary and just return the number?

cmd/metal-api/internal/scaler/poolscaler.go
Gerrit91: Okay, understood. Thanks.

cmd/metal-api/internal/service/v1/partition.go
```diff
@@ -18,13 +18,17 @@ type PartitionBootConfiguration struct {
 type PartitionCreateRequest struct {
 	Common
 	PartitionBase
-	PartitionBootConfiguration PartitionBootConfiguration `json:"bootconfig" description:"the boot configuration of this partition"`
+	PartitionBootConfiguration  PartitionBootConfiguration `json:"bootconfig" description:"the boot configuration of this partition"`
+	PartitionWaitingPoolMinSize *string                    `json:"waitingpoolminsize" description:"the minimum waiting pool size of this partition"`
+	PartitionWaitingPoolMaxSize *string                    `json:"waitingpoolmaxsize" description:"the maximum waiting pool size of this partition"`
```
Gerrit91, line 23: Can be put into PartitionBase and should add `optional:"true"`.

cmd/metal-api/internal/fsm/states/waiting.go
Gerrit91: In the transition we append the event to the container, but why do we need to call the scaler? Isn't the current implementation calling the scaler two times in a row in a matter of microseconds? In this short period of time, the state in the database will not have changed?

iljarotar: If a machine changes from some previous state to waiting, the fsm will call OnTransition but not OnLeave.

Gerrit91: Okay, sorry, I got it now. I think the `OnTransition` triggered the wrong thoughts in my brain. Maybe `OnEnter` would really be better.

cmd/metal-api/internal/scaler/poolscaler_test.go
```diff
@@ -0,0 +1,160 @@
...
+			name: "more waiting machines needed in percent; power on 10 machines",
+			partition: &metal.Partition{
+				WaitingPoolMinSize: "30%",
+				WaitingPoolMaxSize: "40%",
+			},
+			mockFn: func(mock *MockMachineManager) {
+				mock.On("AllMachines").Once().Return(append(metal.Machines{}, make([]metal.Machine, 150)...), nil)
+				mock.On("WaitingMachines").Once().Return(append(metal.Machines{}, make([]metal.Machine, 35)...), nil)
+				mock.On("ShutdownMachines").Once().Return(append(metal.Machines{}, make([]metal.Machine, 20)...), nil)
+				mock.On("PowerOn", &metal.Machine{}).Return(nil).Times(18)
```
Gerrit91, line 65: Isn't 10 machines expected? There is also the possibility to call `manager.AssertExpectations(t)` at the end of a test to make sure the correct amount of calls was made (e.g. when a call was not made).

Gerrit91 (approved):

cmd/metal-api/internal/scaler/poolscaler_test.go
iljarotar: I forgot to update the name of the test. 18 machines are expected:  
```
min = 0.3*150 = 45
max = 0.4*150 = 60
average = round(52.5) = 53
waiting = 35
waiting - average = -18
```

Is calling `manager.AssertExpectations(t)` really necessary? My tests were always failing, if an incorrect number of calls were made.

mwennrich:
As we have different machine types (n1, c1, m1, ...) ... shouldn't we try to have e.g. 10% of every machine in Waiting. As far as I can see the current approach relies on randomness, that we really have enough waiting machines of every machine type?

Gerrit91:
Thanks for taking a look! :) 

All queries are scoped considering the partition and size (as can be seen in `poolsize.go`, which is part providing the data for the `PoolScaler`). So in every partition we ensure that every machine type is within the given user range.

cmd/metal-api/internal/datastore/poolsize.go
```diff
@@ -0,0 +1,105 @@
...
+	return shutdownMachines, nil
+}
+
+func (m *manager) Shutdown(machine *metal.Machine) error {
+	state := metal.MachineState{
+		Value:       metal.ShutdownState,
+		Description: "shut down as exceeding maximum partition poolsize",
+	}
+
+	err := m.rs.publishCommandAndUpdate(m.rs.log, machine, m.publisher, metal.MachineOnCmd, state)
```
Gerrit91, line 74: Has to be `MachineOffCmd`.

Gerrit91 (changes requested):
Found the following work items when rolling this out to our test environment:

- [ ] Machines are not shut down (see code comment)
- [ ] Alive events during wait are triggering the scaler as well, self transitions should not trigger the scaler
- [ ] Log fields for calculated min and max + current value would be great
- [ ] For some reason two machines (expected 1) were scaled down right at the start, see log:

   ```
   metal-api-59479c98db-rz5k5 metal-api {"level":"info","timestamp":"2023-01-13T12:08:07Z","logger":"event-service.pool-scaler","caller":"scaler/poolscaler.go:66","msg":"shut down spare machines","partition":"fra-equ01","number":1}          
   metal-api-59479c98db-rz5k5 metal-api {"level":"info","timestamp":"2023-01-13T12:08:10Z","logger":"event-service.pool-scaler","caller":"scaler/poolscaler.go:66","msg":"shut down spare machines","partition":"fra-equ01","number":1}          
   metal-api-59479c98db-rz5k5 metal-api {"level":"info","timestamp":"2023-01-13T12:08:27Z","logger":"event-service.pool-scaler","caller":"scaler/poolscaler.go:61","msg":"pool size condition met; doing nothing","partition":"fra-equ01"}       
   metal-api-59479c98db-rz5k5 metal-api {"level":"info","timestamp":"2023-01-13T12:08:27Z","logger":"event-service.pool-scaler","caller":"scaler/poolscaler.go:61","msg":"pool size condition met; doing nothing","partition":"fra-equ01"}
   ```

cmd/metal-api/internal/datastore/machine.go
```diff
@@ -21,6 +21,7 @@ type MachineSearchQuery struct {
 	Tags        []string `json:"tags" optional:"true"`
 
 	// allocation
+	NotAllocated        *bool       `json:"not_allocated" optional:"true"`
```
Gerrit91, line 24: This can now be integration-tested fairly easily.

cmd/metal-api/internal/datastore/machine.go
```diff
@@ -53,6 +54,7 @@ type MachineSearchQuery struct {
 	DiskSizes []int64  `json:"disk_sizes" optional:"true"`
 
 	// state
+	Sleeping   *bool   `json:"sleeping" optional:"true"`
```
Gerrit91, line 57: Maybe `Hibernated` would also be a good term to describe what's happening.

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -1771,6 +1772,10 @@ func MachineLiveliness(ds *datastore.RethinkStore, logger *zap.SugaredLogger) er
 }
 
 func evaluateMachineLiveliness(ds *datastore.RethinkStore, m metal.Machine) (metal.MachineLiveliness, error) {
+	if m.State.Sleeping {
+		return metal.MachineLivelinessAlive, nil
```
Gerrit91, line 1776: Maybe we can introduce `metal.MachineLivelinessHibernated` to express a hibernated machine.

cmd/metal-api/internal/datastore/poolsize.go
```diff
@@ -0,0 +1,105 @@
...
+		return nil, err
+	}
+
+	return shutdownMachines, nil
+}
+
+func (m *manager) Shutdown(machine *metal.Machine) error {
+	state := metal.MachineState{
+		Sleeping:    true,
+		Description: "shut down as exceeding maximum partition poolsize",
```
Gerrit91, line 71: The field was probably intended for a different purpose. Maybe we can introduce an individual struct like:

```golang
type Hibernation struct {
    Enabled bool
    Description string 
    Timestamp *time.Time
}
```

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -2020,6 +2029,8 @@ func (r *machineResource) machineCmd(cmd metal.MachineCommand, request *restful.
 			Description: description,
 		}
 		needsUpdate = true
+	case metal.MachineOnCmd:
+		newMachine.State = metal.MachineState{Sleeping: false}
```
Gerrit91, line 2033: The actual power state of a machine is already reported by the metal-bmc to the `/machine/ipmi` endpoint. We can there toggle this flag as well.

PR #357, opened 2022-11-10
add long awaited crud services for tenant

majst01:
actually based on 

- [x] https://github.com/metal-stack/masterdata-api/pull/60

Upstream dependends:

- [ ] https://github.com/metal-stack/metalctl/pull/173
- [ ] https://github.com/metal-stack/metal-go/pull/107

Works with the above prs in fra-equ01


Gerrit91:
Looks good already. We should create a PR for metalctl as well and try it out in the mini-lab.

Gerrit91 (approved):

cmd/metal-api/internal/service/tenant-service.go
```diff
@@ -91,3 +133,132 @@ func (r *tenantResource) listTenants(request *restful.Request, response *restful
...
+
+func (r *tenantResource) findTenants(request *restful.Request, response *restful.Response) {
+	var requestPayload v1.TenantFindRequest
+	err := request.ReadEntity(&requestPayload)
+	if err != nil {
+		r.sendError(request, response, httperrors.BadRequest(err))
+		return
+	}
+
+	res, err := r.mdc.Tenant().Find(context.Background(), mapper.ToMdmV1TenantFindRequest(&requestPayload))
```
Gerrit91, line 145: Context from request should be used. Applies for rest of the file.

cmd/metal-api/internal/service/tenant-service.go
```diff
@@ -91,3 +133,132 @@ func (r *tenantResource) listTenants(request *restful.Request, response *restful
...
+		return
+	}
+
+	plr, err := r.mdc.Project().Find(context.Background(), &mdmv1.ProjectFindRequest{TenantId: wrapperspb.String(id)})
+	if err != nil {
+		r.sendError(request, response, defaultError(err))
+		return
+	}
+	if len(plr.Projects) > 0 {
+		r.sendError(request, response, httperrors.BadRequest(errors.New("there are still projects allocated by this tenant")))
```
Gerrit91, line 206: Is not really a bad request, should return default 422 unprocessable entity.

cmd/metal-api/internal/service/tenant-service.go
```diff
@@ -91,3 +133,132 @@ func (r *tenantResource) listTenants(request *restful.Request, response *restful
...
+	existingTenant, err := r.mdc.Tenant().Get(context.Background(), &mdmv1.TenantGetRequest{Id: requestPayload.Tenant.Meta.Id})
+	if err != nil {
+		r.sendError(request, response, defaultError(err))
+		return
+	}
+
+	// new data
+	tenantUpdateData := mapper.ToMdmV1Tenant(&requestPayload.Tenant)
+	// created date is not updateable
+	tenantUpdateData.Meta.CreatedTime = existingTenant.Tenant.Meta.CreatedTime
```
Gerrit91, line 249: From my perspective either the server side should deal with the case when the `CreatedTime` field is empty and apply this default. Or the server requires to always send all fields on updates but then metalctl needs to take care of this.

majst01:
Maybe we should also add a `tenant history` endpoint, together with a `--version` argument to `edit/describe`

Gerrit91 (approved):

PR #373, opened 2023-01-02
Enrich registerswitch with details about OS and management access

majst01:
This will enable us to add more info in the `metalctl switch ls` output to show which version and vendor of the switch operating system is running. Also enables us to implement something like `metalctl switch ssh <name of the switch>` to easy management of switches.


cmd/metal-api/internal/metal/switch.go
```diff
@@ -17,6 +17,15 @@ type Switch struct {
 	Mode               SwitchMode    `rethinkdb:"mode" json:"mode"`
 	LastSync           *SwitchSync   `rethinkdb:"last_sync" json:"last_sync"`
 	LastSyncError      *SwitchSync   `rethinkdb:"last_sync_error" json:"last_sync_error"`
+	OS                 *SwitchOS     `rethinkdb:"os" json:"os"`
+	ManagementIP       string        `rethinkdb:"management_ip" json:"management_ip"`
+	ManagementUser     string        `rethinkdb:"management_user" json:"management_user"`
+	ConsoleCommand     string        `rethinkdb:"console_command" json:"console_command"`
```
GrigoriyMikhalkin, line 23: What `ConsoleCommand` is and what it's for? A little bit unclear.

majst01: It is an optional manually edited field which contains the command to access the console, in our case this could be something like 

```
telnet 10.1.253.254 7009
```

If ConsoleComand is not an emptystring, metalctl switch console <switchname> would then execute this command, otherwise an error is returned which explains what i wrote here.

GrigoriyMikhalkin: And it cannot be reduced to just IP and port?

majst01: No, because console access depends on the type of OutOfBand Infrastructure you have deployed in the DC. Some Datacenters have for example proprietary solutions where you have to use a Windows application or a Java base GUI to have access to the serial interface.

We have a standart console server which only offers `telnet`, but maybe `ssh` in the future. I could not imagine howto implement this in a type safe way.

GrigoriyMikhalkin (approved):
lgtm

PR #416, opened 2023-02-22
Add switch OS, switch search queries and better datastore tests.

Gerrit91:
- takes out unrelated parts of adding switch OS from #351 
- adds additional search queries for switches allowing more efficient filtering in `metalctl`
- much better testing in the datastore layer for switches using a real rethinkdb as a backend


cmd/metal-api/main.go
```diff
@@ -938,7 +938,7 @@ func run() error {
 	// if we will add support for api-tokens as headers, we had to add them
 	// here to. note: the token's should not contain the product (aka. metal)
 	// because customers should have ONE token for many products.
-	// ExposeHeaders:  []string{"X-FITS-TOKEN"},
+	// ExposeHeaders:  []string{"X-TOKEN"},
```
majst01, line 941: unrelated

go.mod
```diff
@@ -190,3 +190,5 @@ require (
 	nhooyr.io/websocket v1.8.7 // indirect
 	tailscale.com v1.32.3 // indirect
 )
+
+require gopkg.in/inf.v0 v0.9.1 // indirect
```
majst01, line 194: wrong block

majst01 (commented):
Like it, but then we must first test this with released metal-hammer and metal-core

cmd/metal-api/internal/datastore/rethinkdb_integration_test.go
```diff
@@ -5,43 +5,61 @@ package datastore
...
-	"go.uber.org/zap/zaptest"
+	"github.com/testcontainers/testcontainers-go"
+	"go.uber.org/zap"
 
 	"testing"
-
-	"github.com/stretchr/testify/require"
 )
 
-func TestRethinkStore_ConflictIsReturned(t *testing.T) {
```
Gerrit91, line 18: This is now already covered in a switch test and will also be covered in all other tests if we start to test like this for the rest of the package. Hence removed.

cmd/metal-api/internal/datastore/rethinkdb_test.go
```diff
@@ -85,31 +85,6 @@ func TestRethinkStore_db(t *testing.T) {
 	}
 }
 
-func TestRethinkStore_Mock(t *testing.T) {
```
Gerrit91, line 88: This test is rather useless and also the linter does not like it, therefore removed.

Gerrit91:
Works on the mini-lab (switch registration plus machine provisioning), so it's compatible with the current metal-core and metal-hammer. No breaking API changes included.

majst01 (approved):

PR #424, opened 2023-02-27
Update go-restful to v3.10, potentially breaking for clients.

Gerrit91:
References https://github.com/emicklei/go-restful/pull/522.


majst01:
v3.11 of go-restful no went back to original behavior

PR #426, opened 2023-03-02
MEP-12: Rack spreading

iljarotar:
This is a first attempt to implement the rack spreading placement strategy as described in https://github.com/metal-stack/docs/pull/125

The algorithm
1. Groups all machines of a project by racks.
2. Groups all machines of a project, that share at least one tag with the tags of the allocation request, by racks.
3. Creates two sets of racks: those preferred by the tags and those preferred by the project
4. Creates a third set of racks by intersecting the previous two.
5. Checks if there are any spare machines left in one of the three resulting sets of racks.
6. Chooses a random machine either from one of the elected racks or from all machines, if the preferred racks have no spare machines left.

The priority of the election criteria is
1. intersection of tags and project
2. tags only
3. project only


cmd/metal-api/internal/datastore/machine.go
```diff
@@ -487,3 +492,121 @@ func (rs *RethinkStore) FindWaitingMachine(partitionid, sizeid string) (*metal.M
...
+	winners := intersect(tagWinners, projectWinners)
+
+	candidates := allMachines
+	if c := flatten(filter(allRacks, winners...)); len(c) > 0 {
+		candidates = c
+	} else if c := flatten(filter(allRacks, tagWinners...)); len(c) > 0 {
+		candidates = c
+	} else if c := flatten(filter(allRacks, projectWinners...)); len(c) > 0 {
+		candidates = c
+	}
```
majst01, line 518: This looks like a perfect example for using a functional implementation like: https://github.com/samber/lo

But i am unsure if this is worth adding another dependency.

Gerrit91: Instead of using generic functions we added this to a specific type. Overhead is actually pretty small so we decided it's ok and did not introduce another library.

PR #432, opened 2023-03-09
If waiting machines are gone, ContextCancelledError is thrown, do not…

majst01:
… ignore that


Gerrit91:
I think this looks more accurate, but only @kolsa knows why these lines have been added. 🙈

/cc @iljarotar

PR #443, opened 2023-05-19
RFC add option to skip transitives in findMachines

majst01:
This would potentially speed up the findMachines call which is actually by far the most demanding and slow api call.

This call is currently done from:

# MCM Metal
https://github.com/metal-stack/machine-controller-manager-provider-metal/blob/master/pkg/provider/core.go#L297

This i called every 30sec per cluster

Where it uses the partition transitive to get partitionID like so:
```go
providerID := encodeMachineID(*m.Partition.ID, *m.ID)
```
This should be changed to a direct partition field as string from the Machine object 

# Metal CCM
https://github.com/metal-stack/metal-ccm/blob/master/pkg/resources/metal/machine.go#L43
This is called at least every minute for every worker

Uses also Partition.ID as shown above

and:
```go
hostname := *m.Allocation.Hostname
machineTags[hostname] = m.Tags
```

Maybe we should change PartitionID to be contained directly in the machine Object and remove Partition Object altogether from this response object. Or simply copy over only machine.PartitionID to machineResponse.Partition.ID to keep backward compatibility
Another option would be to Create a new Endpoint which returns smaller Objects

# Analysing where to most time is spent in the metal-api

could be done with the pprof endpoint like so: https://www.benburwell.com/posts/flame-graphs-for-go-with-pprof/

### DIFF
--- a/cmd/metal-api/internal/service/firewall-service.go
+++ b/cmd/metal-api/internal/service/firewall-service.go
@@ -147,7 +147,7 @@ func (r *firewallResource) findFirewall(request *restful.Request, response *rest
 }
 
 func (r *firewallResource) findFirewalls(request *restful.Request, response *restful.Response) {
-	var requestPayload datastore.MachineSearchQuery
+	var requestPayload v1.MachineFindRequest
 	err := request.ReadEntity(&requestPayload)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


majst01:
Closing, we already gained enough performance without breaking changes of this approach

PR #446, opened 2023-06-14
Add new endpoint to update the ssh public keys of a allocated machine

majst01:


cmd/metal-api/internal/service/machine-service.go
```diff
@@ -397,6 +397,18 @@ func (r *machineResource) webService() *restful.WebService {
 		Returns(http.StatusOK, "OK", v1.MachineResponse{}).
 		DefaultReturns("Error", httperrors.HTTPErrorResponse{}))
 
+	ws.Route(ws.POST("/{id}/sshpubkeys").
```
Gerrit91, line 400: We could also just utilize the existing machine update endpoint, which already does things like these in order to keep the API surface smaller.

majst01: done

Gerrit91 (approved):

PR #449, opened 2023-06-29
fix: empty audit bodies in during search

vknabel:
In some cases the response bodies were empty when reflecting them with the metalctl.

Also updates to latest metal-lib to profit from the request id fixes


cmd/metal-api/internal/service/v1/audit.go
```diff
@@ -65,6 +67,13 @@ func NewAuditResponse(e auditing.Entry) *AuditResponse {
 		body = v
 	case []byte:
 		body = string(v)
+	default:
+		b, err := json.Marshal(v)
+		if err == nil {
+			body = string(b)
+		} else {
+			body = fmt.Sprintf("unknown body: %v", err)
```
majst01, line 75: maybe the payload should also be put into the error ?

vknabel: Good point. I'll replace the err with the value itself as it does not add relevant information over the data itself.

majst01 (approved):

PR #450, opened 2023-07-04
Add metal-core version to switch OS details

majst01:
This enables us to show the metal-core version in `metalctl switch ls -o wide` as well

Related:

- [ ] https://github.com/metal-stack/metal-core/pull/98
- [ ] https://github.com/metal-stack/metalctl/pull/199

### DIFF
--- a/cmd/metal-api/internal/metal/switch.go
+++ b/cmd/metal-api/internal/metal/switch.go
@@ -26,8 +26,9 @@ type Switch struct {
 type Switches []Switch
 
 type SwitchOS struct {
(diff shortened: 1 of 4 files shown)


Gerrit91 (approved):
Perfect, only the choice to put the version beneath SwitchOS feels a bit weird. Not a deal breaker.

majst01:
> Perfect, only the choice to put the version beneath SwitchOS feels a bit weird. Not a deal breaker.

I was also thinking about that, but it is basically a part of the SwitchOS.

PR #453, opened 2023-07-12
Store Status of a switch in a separate table

### LINKED ISSUE #263: Save switch status responses in separate table
This events happen very often and they should not interfere with regular switch write operations, which are way more important to proceed.

majst01:
closes #263 

This changes the return type from `SwitchResponse` to `SwitchNotifyResponse`. The caller (metal-core) does not do anything with the response object anyway. The metal-core still works without modifications. The sole purpose was to be able to have a rest unit test with a testable interface.

The state actually stored in the Switch object is deleted and not migrated to the new table.

Performance of the `/v1/metal/switch/notify` request drops from ~35ms -> ~2.5ms in our test environment. This is because we neither fetch, nor store the whole switch object anymore. CPU of metal-api also dropped as a side effect.

image


cmd/metal-api/internal/datastore/switch.go
```diff
@@ -196,3 +196,39 @@ func (rs *RethinkStore) ConnectMachineWithSwitches(m *metal.Machine) error {
 
 	return nil
 }
+
+// GetSwitchStatus get SwitchStatus for a given switch id
+func (rs *RethinkStore) GetSwitchStatus(id string) (*metal.SwitchStatus, error) {
+	rs.log.Infow("get state", "id", id)
```
Gerrit91, line 202: These kinds of logs should be debug only. metal-api primarily logs requests only.

cmd/metal-api/internal/datastore/switch.go
```diff
@@ -196,3 +196,39 @@ func (rs *RethinkStore) ConnectMachineWithSwitches(m *metal.Machine) error {
...
+		createErr := rs.createEntity(rs.switchStatusTable(), state)
+		if err != nil {
+			return nil, createErr
+		}
+		return state, nil
+	}
+	rs.log.Infow("update state", "old", ss, "new", state)
+	err = rs.updateEntity(rs.switchStatusTable(), state, &ss)
+	return state, err
+}
```
Gerrit91, line 234: You can just use upsertEntity (see event insertions). Should reduce this function to a single line.

cmd/metal-api/internal/datastore/switch.go
```diff
@@ -196,3 +196,39 @@ func (rs *RethinkStore) ConnectMachineWithSwitches(m *metal.Machine) error {
...
+func (rs *RethinkStore) GetSwitchStatus(id string) (*metal.SwitchStatus, error) {
+	rs.log.Infow("get state", "id", id)
+	var ss metal.SwitchStatus
+	err := rs.findEntityByID(rs.switchStatusTable(), &ss, id)
+	if err != nil && !metal.IsNotFound(err) {
+		return nil, err
+	}
+	if err != nil && metal.IsNotFound(err) {
+		return &metal.SwitchStatus{}, nil
+	}
```
Gerrit91, line 210: For consistency, I think this kind of error handling should go into the service layer. Other functions of this package, which do similar things, do not return empty responses on non-existent entities.

cmd/metal-api/internal/service/v1/switch.go
```diff
@@ -92,26 +98,30 @@ type SwitchSync struct {
...
 
-func NewSwitchResponse(s *metal.Switch, p *metal.Partition, nics SwitchNics, cons []SwitchConnection) *SwitchResponse {
+func NewSwitchResponse(s *metal.Switch, ss *metal.SwitchStatus, p *metal.Partition, nics SwitchNics, cons []SwitchConnection) *SwitchResponse {
 	if s == nil {
 		return nil
 	}
 
 	var lastSync *SwitchSync
-	if s.LastSync != nil {
+	if ss.LastSync != nil {
```
Gerrit91, line 107: `ss == nil` should also be checked.

cmd/metal-api/internal/service/switch-service.go
```diff
@@ -210,37 +210,35 @@ func (r *switchResource) notifySwitch(request *restful.Request, response *restfu
...
 		r.sendError(request, response, defaultError(err))
 		return
 	}
 
-	resp, err := makeSwitchResponse(s, r.ds)
-	if err != nil {
-		r.sendError(request, response, defaultError(err))
-		return
-	}
+	r.log.Infow("switchservice notify", "state", newSS, "resp", resp)
```
Gerrit91, line 241: Should be debug level.

cmd/metal-api/internal/service/switch-service.go
```diff
@@ -628,8 +626,12 @@ func makeSwitchResponse(s *metal.Switch, ds *datastore.RethinkStore) (*v1.Switch
 
 	nics := makeSwitchNics(s, ips, machines)
 	cons := makeSwitchCons(s)
+	ss, err := ds.GetSwitchStatus(s.ID)
+	if err != nil {
+		return nil, err
+	}
```
Gerrit91, line 632: Should go into `findSwitchReferencedEntites`, which is where the other references entities are gathered.

Gerrit91 (commented):
Very reasonable addition! Left you some comments.

cmd/metal-api/internal/datastore/switch.go
majst01: But upsert does not create the entity if it does not exists, or do i miss something ?

Gerrit91: I think it does, it uses the same rethinkdb function as for creating entities, but does a replace on existing documents.

cmd/metal-api/internal/datastore/switch.go
majst01: done

cmd/metal-api/internal/datastore/switch.go
majst01: done, and fixed the BUG in upsert not setting creation date

cmd/metal-api/internal/datastore/switch.go
majst01: done

cmd/metal-api/internal/service/v1/switch.go
majst01: done

cmd/metal-api/internal/service/switch-service.go
majst01: removed

cmd/metal-api/internal/service/switch-service.go
majst01: done

Gerrit91 (approved):

mwennrich (approved):

Issue #463, opened 2023-09-08
Add Tenant History Command once #357 was merged


majst01:

Pilzington:
I think I changed all parts that needed to be changed.  
I also get an expected result while testing in the minilab:
```
$ /home/alex/Documents/github.com/metal-stack/metalctl/bin/metalctl-linux-amd64 tenant history --id metal-stack --at 1696580971
ID            NAME          DESCRIPTION                      LABELS   ANNOTATIONS 
metal-stack   metal-stack   metal-stack tenant, which is                            
                            provider             
```
What I did not test was if the function of the masterdata really is working. I was not sure how to achieve different history states.

vknabel:
It turns out to be far too complicated to implement

PR #470, opened 2023-09-28
Implement Tenant History Command

Pilzington:
implements #463

### DIFF
--- a/cmd/metal-api/internal/service/tenant-service.go
+++ b/cmd/metal-api/internal/service/tenant-service.go
@@ -8,6 +8,7 @@ import (
 	v1 "github.com/metal-stack/masterdata-api/api/rest/v1"
 	mdmv1 "github.com/metal-stack/masterdata-api/api/v1"
 	mdm "github.com/metal-stack/masterdata-api/pkg/client"
+	internalV1 "github.com/metal-stack/metal-api/cmd/metal-api/internal/service/v1"
 	"go.uber.org/zap"
 	"google.golang.org/protobuf/types/known/wrapperspb"
 
@@ -70,6 +71,17 @@ func (r *tenantResource) webService() *restful.WebService {
 		Reads(v1.TenantFindRequest{}).
 		Writes([]v1.TenantResponse{}).
 		Returns(http.StatusOK, "OK", []v1.TenantResponse{}).
+		DefaultReturns("Error", httperrors.HTTPErrorResponse{}))
+
+	ws.Route(ws.POST("/{id}/history").
+		To(viewer(r.getTenantHistory)).
+		Operation("getTenantHistory").
+		Doc("get tenant with this id at the given timestamp").
+		Param(ws.PathParameter("id", "identifier of the tenant").DataType("string")).
+		Metadata(restfulspec.KeyOpenAPITags, tags).
+		Reads(internalV1.TenantGetHistoryRequest{}).
+		Writes(v1.TenantResponse{}).
+		Returns(http.StatusOK, "OK", v1.TenantResponse{}).
 		DefaultReturns("Error", httperrors.HTTPErrorResponse{}))
 
 	ws.Route(ws.DELETE("/{id}").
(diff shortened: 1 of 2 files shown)


cmd/metal-api/internal/service/tenant-service.go
```diff
@@ -72,6 +73,17 @@ func (r *tenantResource) webService() *restful.WebService {
 		Returns(http.StatusOK, "OK", []v1.TenantResponse{}).
 		DefaultReturns("Error", httperrors.HTTPErrorResponse{}))
 
+	ws.Route(ws.POST("/{id}/history").
+		To(viewer(r.getTenantHistory)).
+		Operation("getTenantHistory").
+		Doc("get tenant with this id at the given timestamp").
+		Param(ws.PathParameter("id", "identifier of the tenant").DataType("string")).
+		Metadata(restfulspec.KeyOpenAPITags, tags).
```
Gerrit91, line 81: Maybe it makes sense to exclude this from auditing as this is just a read operation:

```suggestion
		Metadata(restfulspec.KeyOpenAPITags, tags).
		Metadata(auditing.Exclude, true).
```

cmd/metal-api/internal/service/v1/tenant.go
```diff
@@ -0,0 +1,9 @@
+package v1
+
+import "google.golang.org/protobuf/types/known/timestamppb"
+
+type (
+	TenantGetHistoryRequest struct {
+		At *timestamppb.Timestamp
```
Gerrit91, line 7: Timestamppb is for protobuf. For this API you should just use time.Time.

Explicit json annotation + description would be great as this will be rendered into the swagger spec. Also you need to check in the generated swagger spec.

cmd/metal-api/internal/service/tenant-service.go
```diff
@@ -8,6 +8,7 @@ import (
 	v1 "github.com/metal-stack/masterdata-api/api/rest/v1"
 	mdmv1 "github.com/metal-stack/masterdata-api/api/v1"
 	mdm "github.com/metal-stack/masterdata-api/pkg/client"
+	internalV1 "github.com/metal-stack/metal-api/cmd/metal-api/internal/service/v1"
```
Gerrit91, line 11: Maybe just call this API v1 as in the other packages and rename the existing v1 import to mdmv1rest or something.

cmd/metal-api/internal/service/tenant-service.go
```diff
@@ -72,6 +73,17 @@ func (r *tenantResource) webService() *restful.WebService {
 		Returns(http.StatusOK, "OK", []v1.TenantResponse{}).
 		DefaultReturns("Error", httperrors.HTTPErrorResponse{}))
 
+	ws.Route(ws.POST("/{id}/history").
+		To(viewer(r.getTenantHistory)).
+		Operation("getTenantHistory").
+		Doc("get tenant with this id at the given timestamp").
+		Param(ws.PathParameter("id", "identifier of the tenant").DataType("string")).
+		Metadata(restfulspec.KeyOpenAPITags, tags).
+		Reads(internalV1.TenantGetHistoryRequest{}).
```
Pilzington, line 82: @Gerrit91  Would it be possible to just reuse [the original](https://github.com/metal-stack/masterdata-api/blob/master/api/v1/tenant.pb.go#L300), or would that be a bad pratice?

Gerrit91: We should not mix the GRPC APIs with the REST APIs.

Gerrit91: What you could do is adding the request to the REST definition in the masterdata-api: https://github.com/metal-stack/masterdata-api/blob/master/api/rest/v1/tenant.go

Gerrit91:
This is the follow-up issue: https://github.com/metal-stack/masterdata-api/issues/79

vknabel:
Not planned to go down this road, see https://github.com/metal-stack/metal-api/issues/463

PR #471, opened 2023-10-02
Add endpoint for machine issues.

### LINKED ISSUE #466: Move machine issue evaluation from metalctl to metal-api
Replaces https://github.com/metal-stack/metalctl/issues/22.

Original issue description:

> I think this can lead to a bad user experience that metalctl does not use the same code for evaluating if a machine has issues that metal-api uses for partition capacity. Both implement their own set of conditions.

So, in this issue we should:

- Move the machine issue evaluation to the metal-api
- Create a new metal-api endpoint v1/machine/issues and return a dedicated machine issue response to the client
- metalctl should not need to implement any custom logic anymore
- Use this same machine issue evaluation for counting faulty machines in partition capacity

Gerrit91:
Closes #466.

### DIFF
--- a/cmd/metal-api/internal/issues/asn-uniqueness.go
+++ b/cmd/metal-api/internal/issues/asn-uniqueness.go
@@ -1,0 +1,122 @@
+package issues
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+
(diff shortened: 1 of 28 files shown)


majst01:
Nice enhancement ! Had some naming issues, but already fixed.

majst01 (approved):

PR #484, opened 2023-12-28
Introduce size reservations for projects.

Gerrit91:
I think we once discussed this in the past but for some reason we said it's impossible to implement. I can't remember why anymore, so here is a suggestion.

This allows reserving machines of a specific size for projects in the the allocation process.

This is feature can be important for the operators to have some spare machines available, e.g. for updating Gardener Seed clusters while preventing regular users to allocate the last machines available in a partition. There are also use-cases where certain customers want to use a certain machine size exclusively, which can also be realized with this approach.

Unfortunately, just like the rack spreading, this algorithm is best-effort only. It can be tricked by heavy parallelization because we do not run the `findMachineCandidate` function in sync.

### DIFF
--- a/cmd/metal-api/internal/datastore/machine.go
+++ b/cmd/metal-api/internal/datastore/machine.go
@@ -428,12 +428,12 @@ func (rs *RethinkStore) UpdateMachine(oldMachine *metal.Machine, newMachine *met
 // FindWaitingMachine returns an available, not allocated, waiting and alive machine of given size within the given partition.
 // TODO: the algorithm can be optimized / shortened by using a rethinkdb join command and then using .Sample(1)
 // but current implementation should have a slightly better readability.
-func (rs *RethinkStore) FindWaitingMachine(projectid, partitionid, sizeid string, placementTags []string) (*metal.Machine, error) {
+func (rs *RethinkStore) FindWaitingMachine(projectid, partitionid string, size metal.Size, placementTags []string) (*metal.Machine, error) {
 	q := *rs.machineTable()
 	q = q.Filter(map[string]interface{}{
 		"allocation":  nil,
 		"partitionid": partitionid,
-		"sizeid":      sizeid,
+		"sizeid":      size.ID,
 		"state": map[string]string{
 			"value": string(metal.AvailableState),
 		},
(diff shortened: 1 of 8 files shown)


cmd/metal-api/internal/datastore/machine.go
```diff
@@ -499,6 +503,43 @@ func (rs *RethinkStore) FindWaitingMachine(projectid, partitionid, sizeid string
...
+// checkSizeReservations returns true when an allocation is possible and
+// false when size reservations prevent the allocation for the given project in the given partition
+func checkSizeReservations(available metal.Machines, projectid, partitionid string, machinesByProject map[string]metal.Machines, size metal.Size) bool {
+	if size.Reservations == nil {
+		return true
+	}
+
+	var (
+		reservations = 0
+		max          = func(x, y int) int {
```
majst01, line 515: There is already min/max available in go since >= 1.21, see: https://tip.golang.org/ref/spec#Min_and_max

cmd/metal-api/internal/metal/size.go
```diff
@@ -138,6 +157,33 @@ func (s *Size) Overlaps(ss *Sizes) *Size {
...
+	for _, r := range rs {
+		r := r
+		if slices.Contains(r.PartitionIDs, partitionID) {
+			result = append(result, r)
+		}
+	}
+	return result
+}
+
+func (rs Reservations) Validate() error {
```
majst01, line 171: There must be also a validation for duplicate partitionIDs

majst01 (changes requested):
Also if project is deleted, the reservation must be deleted

Gerrit91:
References https://github.com/metal-stack/metal-api/pull/134.

cmd/metal-api/internal/datastore/machine.go
```diff
@@ -499,6 +503,43 @@ func (rs *RethinkStore) FindWaitingMachine(projectid, partitionid, sizeid string
...
+		reservations += r.Amount
+
+		alreadyAllocated := len(machinesByProject[r.ProjectID])
+
+		if projectid == r.ProjectID && alreadyAllocated < r.Amount {
+			// allow allocation for the project when it has a reservation and there are still allocations left
+			return true
+		}
+
+		// substract aleady used up reservations of the project
```
majst01, line 536:
```suggestion
		// substract already used up reservations of the project
```

majst01 (approved):

PR #488, opened 2024-01-10
Change License to Apache-2

majst01:
We had AGPL because we wanted to enforce third parties which eventually use also metal-stack to contribute back their enhancements. We did not get any external contributions to metal-api in the past. AGPL is also a forbidden LICENSE for a lot of companies to contribute to which might be the reason why we did not get any. Other repos in our org, which are Apache-2.0 or MIT licensed, did get quite a lot of contributions.

I added a lot of reviewers which contributed in the past to let everyone know about my proposed change.


Gerrit91 (approved):

vknabel (approved):

robertvolkmann:
I agree to the license change.

robertvolkmann (approved):

chbmuc:
We chose the AGPL to make sure that anyone else using and improving the core components, has to contribute back. I'm not convinced that a more liberal license would really foster the use of metal-stack in a way that we get more contributions, even though they don't need to contribute any longer.

mreiger:
I never understood all the FUD against copyleft licenses...

PR #490, opened 2024-01-18
Prevent parallel allocation to ensure spreading and reservations.

Gerrit91:
This implementation implements a lock / mutex through the RethinkDB. I hope the tests I wrote are good enough to build trust into this idea as I think the advantages of having spread and reservations truly work are pretty big. If someone has a better idea to achieve the same outcome, please let me know.


Gerrit91:
Possible optimization would be to lock with the partition key such that allocations can go in parallel in different partitions.

ulrichSchreiner (approved):

cmd/metal-api/internal/datastore/shared_mutex.go
```diff
@@ -0,0 +1,194 @@
...
+
+const (
+	// defaultSharedMutexCheckInterval is the interval in which it checked whether mutexes have expired. if they have expired, they will be released.
+	// this is a safety mechanism in case a mutex was forgotten to be released to prevent the whole machinery to lock up forever.
+	defaultSharedMutexCheckInterval = 30 * time.Second
+	// defaultSharedMutexAcquireTimeout defines a timeout for the context for the acquisition of the mutex.
+	defaultSharedMutexAcquireTimeout = 10 * time.Second
+)
+
+type mutexOpt interface{}
```
majst01, line 38:
```suggestion
type mutexOpt any
```

cmd/metal-api/internal/datastore/shared_mutex.go
```diff
@@ -0,0 +1,194 @@
...
+		table:         table,
+		checkinterval: timeout,
+	}
+
+	go m.expireloop(ctx)
+
+	return m, nil
+}
+
+type lockOpt interface{}
```
majst01, line 73:
```suggestion
type lockOpt any
```

majst01 (approved):

PR #492, opened 2024-01-22
Add endpoints to manage switch ports.

Gerrit91:
This would be extremely helpful for integration testing BGP failover scenarios.


Gerrit91:
Is "Quark".

PR #506, opened 2024-02-26
allow the API to toggle switch ports

ulrichSchreiner:
This API enables the user to toggle a port of a switch. The endpoint is `/switch/{id}/port` and the post body contains 
~~~go
type SwitchPortToggleRequest struct {
	NicName string           `json:"nic" description:"the nic of the switch you want to change"`
	Status  SwitchPortStatus `json:"status" description:"sets the port status" enum:"UP|DOWN"`
}
~~~

When this endpoint is invoked, the system stores the `status` as the new desired state (if it differs from the current state). The switch uses [metal-core](https://github.com/metal-stack/metal-core) to regularly fetch the switch state. If it finds a desired state which differs from the current state, it will change the switch configuration to reflect this state and send a notification to `/switch/{id}/notify` and report the new state. If the new state is equal to the desired state, **metal-api** will store this new state and remove the desired state from the persistence layer.

Dependencies:

 - [ ] https://github.com/metal-stack/metal-core/pull/119
 - [ ] https://github.com/metal-stack/metalctl/pull/241


cmd/metal-api/internal/service/switch-service.go
```diff
@@ -313,11 +315,18 @@ func (r *switchResource) toggleSwitchPort(request *restful.Request, response *re
...
+
+	// Updates the state of a NIC on the switch if the requested state change is valid
+	//
+	// Loops through each NIC on the switch and checks if the name matches the
+	// requested NIC. If a match is found, it tries to update the NIC's state to
+	// the requested state. If the state change is valid, it sets the new state on
+	// the NIC and marks that an update was made.
 	for i, nic := range newSwitch.Nics {
-		if nic.Name == nicname {
+		// compre nic-names case-insensitive
```
majst01, line 328: compare ?

.golangci.yaml
```diff
@@ -3,4 +3,8 @@ run:
   deadline: 10m
 linters:
   disable:
-    - musttag
\ No newline at end of file
+    - musttag
+  enable:
+    - testifylint
+    - unused
```
majst01, line 9: we also check for bugs

majst01:
We should run the small integration test suite with this deployed in test, then merge

cmd/metal-api/internal/service/switch-service.go
```diff
@@ -244,9 +256,107 @@ func (r *switchResource) notifySwitch(request *restful.Request, response *restfu
...
+	// the NIC and marks that an update was made.
+	for i, nic := range newSwitch.Nics {
+		// compare nic-names case-insensitive
+		if strings.EqualFold(nic.Name, requestPayload.NicName) {
+			found = true
+			newstate, changed := nic.State.WantState(desired)
+			if changed {
+				newSwitch.Nics[i].State = &newstate
+				updated = true
+				break
```
Gerrit91, line 335: Could be placed after the if-condition because the iteration can always stop as soon as the nic was found.

cmd/metal-api/internal/metal/network.go
```diff
@@ -18,6 +45,103 @@ type Nic struct {
...
+				Actual:  s,
+				Desired: nil,
+			}, true
+		}
+		return *ns, false
+	}
+	return NicState{
+		Actual:  ns.Actual,
+		Desired: &s,
+	}, true
```
Gerrit91, line 144: Hope I am not mistaken, but I think it's not covering all cases, for example let's assume the following nic state:

``` 
{"actual": "UP", "desired:" "DOWN"}
```

The user then sends `DOWN` a second time, which will lead to the right return but it will indicate a changed state with `true` but it actually did not change anything.

Gerrit91 (approved):

PR #508, opened 2024-02-28
remove dead code

ulrichSchreiner:
remove dead code.

in the image-service the result of `ListMachines` is only used later when doing the lookup via `machinesByImage`. But after this lookup we have a nested if-construct where the inner "if" is always false as it was checked a few lines above. so the `if len(machiens) ...` is alway an empty block and can be removed. so the whole `Listmachines` .

same in `network-service` as the `if err != nil` is always false.

### DIFF
--- a/cmd/metal-api/internal/service/image-service.go
+++ b/cmd/metal-api/internal/service/image-service.go
@@ -367,20 +367,6 @@ func (r *imageResource) deleteImage(request *restful.Request, response *restful.
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


Gerrit91:
Great that you found this. Can't we just remove the inner if-clauses that check the error to make the code behave as it was intended?

ulrichSchreiner:
ok, code changed so it behaves like the author originally wanted

majst01 (approved):

PR #515, opened 2024-03-21
Validate machine response against swagger spec.

Gerrit91:
We often produce responses with our server that are not valid against the Swagger spec. For Go clients this seems not to be a big issue as type safety is guaranteed and it's not possible to produce server responses that violate type constraints. For other clients like Python there are much stricter validations though.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/emicklei/go-restful-openapi/v2 v2.9.1
 	github.com/emicklei/go-restful/v3 v3.11.3
 	github.com/go-openapi/spec v0.21.0
+	github.com/go-openapi/validate v0.24.0
 	github.com/google/go-cmp v0.6.0
 	github.com/google/uuid v1.6.0
 	github.com/grpc-ecosystem/go-grpc-middleware/providers/prometheus v1.0.0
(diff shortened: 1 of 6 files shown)


go.mod
```diff
@@ -77,11 +78,13 @@ require (
...
 	github.com/go-logr/stdr v1.2.2 // indirect
 	github.com/go-ole/go-ole v1.3.0 // indirect
+	github.com/go-openapi/analysis v0.23.0 // indirect
 	github.com/go-openapi/errors v0.22.0 // indirect
 	github.com/go-openapi/jsonpointer v0.21.0 // indirect
 	github.com/go-openapi/jsonreference v0.21.0 // indirect
+	github.com/go-openapi/loads v0.22.0 // indirect
 	github.com/go-openapi/runtime v0.28.0 // indirect
-	github.com/go-openapi/strfmt v0.23.0 // indirect
+	github.com/go-openapi/strfmt v0.23.0
```
majst01, line 87: should be in the require section

Gerrit91: Good catch. Resolved.

Gerrit91:
For some fields we were lucky before because we returned directly what was returned from the RethinkDB, which never returns `nil` but an empty slice for a document. Therefore we fulfilled the spec's contract, e.g. for `destinationprefixes`.

PR #522, opened 2024-04-30
Migrate ipam to use the grpc service

majst01:
With this PR we move from library usage of go-ipam to a dedicated grpc go-ipam service. This removes the direct dependency to the ipam database as well and makes the overall architecture clearer.

TODO:

- [x] https://github.com/metal-stack/go-ipam/pull/86
- [ ] https://github.com/metal-stack/helm-charts/pull/40
- [ ] https://github.com/metal-stack/metal-roles/pull/99
- [ ] https://github.com/metal-stack/mini-lab/pull/168
- [x] update deployment diagram in docs
- [x] check for notfound error raised and handled on ip/prefix delete and alreadyexists

Replaces #312


cmd/metal-api/main.go
```diff
@@ -738,7 +742,7 @@ func initRestServices(audit auditing.Auditing, withauth bool, ipmiSuperUser meta
 		log.Fatal(err)
 	}
 
-	healthService, err := rest.NewHealth(logger, service.BasePath, ds) // FIXME
+	healthService, err := rest.NewHealth(logger, service.BasePath, ds, ipamer) // FIXME
```
Gerrit91, line 745: FIXME can be removed?

cmd/metal-api/main.go
```diff
@@ -817,6 +821,9 @@ func initHeadscale() error {
 }
 
 func dumpSwaggerJSON() {
+
+	// This is total fuckup
```
Gerrit91, line 825: No swearing.

Gerrit91 (commented):
Almost done.

cmd/metal-api/internal/datastore/network.go
```diff
@@ -91,39 +91,45 @@ func (p *NetworkSearchQuery) generateTerm(rs *RethinkStore) *r.Term {
 	}
 
 	for _, prefix := range p.Prefixes {
-		ip, length := utils.SplitCIDR(prefix)
+		pfx, err := netip.ParsePrefix(prefix)
+		if err != nil {
+			continue
```
vknabel, line 96: could these `continue`s lead to a different behavior? Previously it would set `ip` to be `prefix`. Now it will be ignored.
Or are they forbidden and validated beforehand?

majst01: We should error out then, will do

majst01: done

vknabel: Perfect! On line 117 there is another `continue`. Didn't want to open two conversations.

vknabel (approved):

mwennrich (approved):

cmd/metal-api/main.go
```diff
@@ -553,30 +558,31 @@ func initMasterData() {
...
 func initIpam() {
-	dbAdapter := viper.GetString("ipam-db")
-	switch dbAdapter {
-	case "postgres":
-		pgStorage, err := goipam.NewPostgresStorage(
-			viper.GetString("ipam-db-addr"),
-			viper.GetString("ipam-db-port"),
-			viper.GetString("ipam-db-user"),
-			viper.GetString("ipam-db-password"),
-			viper.GetString("ipam-db-name"),
```
Gerrit91, line 564: Removed flags need to be removed from the cobra's root command as well. The new flag should be added to the root command.

cmd/metal-api/internal/datastore/network.go
```diff
@@ -91,42 +92,48 @@ func (p *NetworkSearchQuery) generateTerm(rs *RethinkStore) *r.Term {
 	}
 
 	for _, prefix := range p.Prefixes {
-		ip, length := utils.SplitCIDR(prefix)
+		pfx, err := netip.ParsePrefix(prefix)
+		if err != nil {
+			return nil, fmt.Errorf("unable to parse prefix %w", err)
```
Gerrit91, line 97: It'd be nice if we could validate this in the service layer in order to properly return `BadRequest` status code.

cmd/metal-api/internal/datastore/network.go
```diff
@@ -91,42 +92,48 @@ func (p *NetworkSearchQuery) generateTerm(rs *RethinkStore) *r.Term {
...
+		}
+		ip := pfx.Addr()
+		length := pfx.Bits()
 
 		q = q.Filter(func(row r.Term) r.Term {
 			return row.Field("prefixes").Map(func(p r.Term) r.Term {
 				return p.Field("ip")
-			}).Contains(r.Expr(ip))
+			}).Contains(r.Expr(ip.String()))
```
Gerrit91, line 105: I think in a follow-up PR we should write an integration test (similar to how it was done for machine entity) to verify that all these filters are really working. I can do this after this was merged.

cmd/metal-api/main.go
majst01: Done

cmd/metal-api/internal/datastore/network.go
majst01: BadRequest is not return from the service layer

cmd/metal-api/internal/datastore/network.go
majst01: We had one, but struggled to make this work in the integration test in CI

cmd/metal-api/main.go
```diff
@@ -244,13 +247,6 @@ func init() {
 	rootCmd.PersistentFlags().StringP("db-user", "", "", "the database user to use")
 	rootCmd.PersistentFlags().StringP("db-password", "", "", "the database password to use")
 
-	rootCmd.Flags().StringP("ipam-db", "", "postgres", "the database adapter to use")
```
Gerrit91, line 247:
```golang
	rootCmd.Flags().String("ipam-grpc-server-endpoint", "", "the ipam grpc server endpoint")
```

majst01: Sure :-) done

cmd/metal-api/internal/datastore/network.go
Gerrit91: https://github.com/metal-stack/metal-api/pull/541

PR #526, opened 2024-05-21
Allow changing switch OS during replace

iljarotar:
This PR maps Cumulus port names to SONiC port names to allow twin switches to run different OSs.


cmd/metal-api/internal/utils/utils.go
```diff
@@ -41,3 +41,15 @@ func GetOsAndSemverFromImage(id string) (string, *semver.Version, error) {
 	}
 	return os, v, nil
 }
+
+func ErrorsAreEqual(err1, err2 error) bool {
```
majst01, line 45: only used in test, please unexport and put into the test files where used

iljarotar:
replaced by https://github.com/metal-stack/metal-api/pull/566

PR #527, opened 2024-05-27
Update go, alpine, proto toolchain

majst01:


mwennrich (approved):

cmd/metal-api/internal/datastore/shared_mutex.go
```diff
@@ -76,7 +76,7 @@ type lockOptAcquireTimeout struct {
 	timeout time.Duration
 }
 
-func newLockOptAcquireTimeout(t time.Duration) *lockOptAcquireTimeout {
+func newLockOptAcquireTimeout(t time.Duration) *lockOptAcquireTimeout { // nolint:unused
```
Gerrit91, line 79: It's used in the integration packages which is by default excluded.

ulrichSchreiner (approved):

PR #528, opened 2024-05-28
Support different disk types for size matching

majst01:


mwennrich (approved):

ulrichSchreiner:
what about the cases where 

  - the pattern has a typo and is illegal/wrong
  - you have two constraints with oberlapping patterns (`/dev/nvm*` and `/dev/nvmeß*`)

should we care? or did i failed to see the code?

majst01:
> what about the cases where
> 
> * the pattern has a typo and is illegal/wrong

How should we identify illegal/wrong patterns ? maybe like so:

```go
if _, err := filepath.Match(pattern, ""); err != nil {
		return nil, err
}
```

> * you have two constraints with oberlapping patterns (`/dev/nvm*` and `/dev/nvme0*`)
> 

Also hard to tell ?

> should we care? or did i failed to see the code?

ulrichSchreiner (approved):

Gerrit91 (approved):

PR #535, opened 2024-06-07
Add size reservation filters.

Gerrit91:


cmd/metal-api/internal/service/partition-service.go
```diff
@@ -466,10 +466,9 @@ func (r *partitionResource) calcPartitionCapacity(pcr *v1.PartitionCapacityReque
 			size := sizesByID[cap.Size]
 
 			for _, reservation := range size.Reservations.ForPartition(pc.ID) {
-				reservation := reservation
-
 				cap.Reservations += reservation.Amount
 				cap.UsedReservations += min(len(machinesByProject[reservation.ProjectID].WithSize(size.ID).WithPartition(pc.ID)), reservation.Amount)
+				cap.Free -= cap.UsedReservations
```
Gerrit91, line 471: This is subject of discussion: Adds visibility to "overbooking" a certain size.

cmd/metal-api/internal/datastore/size.go
```diff
@@ -4,8 +4,61 @@ import (
...
 )
 
+// SizeSearchQuery can be used to search sizes.
+type SizeSearchQuery struct {
+	ID                    *string           `json:"id" optional:"true"`
+	Name                  *string           `json:"name" optional:"true"`
+	Labels                map[string]string `json:"labels" optional:"true"`
+	ReservationsPartition *string           `json:"reservations_partition" optional:"true"`
+	ReservationsProject   *string           `json:"reservations_project" optional:"true"`
+}
```
majst01, line 17: This is hard to get, what about putting these into a separate struct like so:

```suggestion
type SizeSearchQuery struct {
	ID                    *string           `json:"id" optional:"true"`
	Name                  *string           `json:"name" optional:"true"`
	Labels                map[string]string `json:"labels" optional:"true"`
        Reservation *Reservation           `json:"reservation" optional:"true"`
}

type Reservation struct {
     Partition *string
     Project *string
}

```

cmd/metal-api/internal/service/partition-service.go
majst01: Will this expose the free machines of a certain size ? if yes i would fill this field only for provider-tenants

cmd/metal-api/internal/service/v1/size.go
```diff
@@ -50,7 +50,14 @@ type SizeReservationResponse struct {
...
 	ProjectAllocations int               `json:"projectallocations" description:"the amount of allocations of this project referenced by this size reservation"`
-	Labels             map[string]string `json:"labels" description:"free labels associated with this size reservation."`
+	Labels             map[string]string `json:"labels,omitempty" description:"free labels associated with this size reservation."`
+}
+
+type SizeReservationListRequest struct {
+	SizeID      *string `json:"sizeid,omitempty" description:"the size id of this size reservation"`
+	Tenant      *string `json:"tenant,omitempty" description:"the tenant of this size reservation"`
+	ProjectID   *string `json:"projectid,omitempty" description:"the project id of this size reservation"`
+	PartitionID *string `json:"partitionid,omitempty" description:"the partition id of this size reservation"`
```
majst01, line 60: Here its a different from the SizeSearchQuery ?

cmd/metal-api/internal/datastore/size.go
Gerrit91: Done.

cmd/metal-api/internal/service/v1/size.go
Gerrit91: Because I wanted to add the tenant, which is an information that the database layer is not aware of. :sweat:

cmd/metal-api/internal/service/partition-service.go
Gerrit91: Moved into #538.

majst01 (approved):

PR #536, opened 2024-06-07
Fix bug on checking headscale connection.

Gerrit91:
This bug can lead to flapping VPN connections for machines in case a machine was already allocated before but through a different project. Because we used a map implementation, the behavior was pretty much non-deterministic. See this content from headscale as an example:

```
❯ k exec -it headscale-cc9bbbd47-xtlbh -- headscale nodes list | grep 5e214c00-837a-11e9-8000-3cecef23001e                                                                                                                                                                                                                                                                             
608 | 5e214c00-837a-11e9-8000-3cecef23001e | 5e214c00-837a-11e9-8000-3cecef23001e-hery3tdd | [c5ISp]    | [0j/5X] | ef755a9e-37ec-4c70-a67e-f0237e799d4a | , fd7a:115c:a1e0:ab12:2000::18 | false     | 2024-06-07 14:03:49 | 0001-01-01 00:00:00 | online  | no                                                             
564 | 5e214c00-837a-11e9-8000-3cecef23001e | 5e214c00-837a-11e9-8000-3cecef23001e          | [e+RXP]    | [AbKaj] | 1eaa31c0-e092-4a3c-b795-5969b7dfd9c9 | , fd7a:115c:a1e0:ab12:2000::11 | false     | 2024-06-03 12:43:34 | 0001-01-01 00:00:00 | offline | no  
```

### DIFF
--- a/cmd/metal-api/main.go
+++ b/cmd/metal-api/main.go
@@ -891,8 +891,7 @@ func evaluateVPNConnected() error {
 		return err
 	}
 
-	ms := metal.Machines{}
-	err = ds.SearchMachines(&datastore.MachineSearchQuery{AllocationRole: &metal.RoleFirewall}, &ms)
+	ms, err := ds.ListMachines()
 	if err != nil {
 		return err
 	}
--- a/cmd/metal-api/main.go
+++ b/cmd/metal-api/main.go
@@ -15,7 +15,6 @@ import (
 	"syscall"
 	"time"
 
-	"golang.org/x/exp/slices"
 	"google.golang.org/protobuf/types/known/wrapperspb"
 
 	"github.com/Masterminds/semver/v3"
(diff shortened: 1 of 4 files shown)


cmd/metal-api/main.go
```diff
@@ -888,15 +891,16 @@ func evaluateVPNConnected() error {
 		return err
 	}
 
-	ms, err := ds.ListMachines()
+	ms := metal.Machines{}
+	err = ds.SearchMachines(&datastore.MachineSearchQuery{AllocationRole: &metal.RoleFirewall}, &ms)
```
majst01, line 895: i think we must search for all machines because there are some machines, not firewalls, which are connected, like storage in test.

Gerrit91: That's a good point.

cmd/metal-api/main.go
```diff
@@ -907,8 +910,25 @@ func evaluateVPNConnected() error {
 		if m.Allocation == nil || m.Allocation.VPN == nil {
 			continue
 		}
-		connected := connectedMap[m.ID]
+
+		connected := slices.ContainsFunc(headscaleMachines, func(headscaleMachine *headscalev1.Machine) bool {
```
majst01, line 914: It would be better if we can keep the headscale dependency in the headscale package.

Gerrit91: I wanted to move this to the service layer similar to the other evaluation commands and in order to be able to write a test. Maybe now it's better.

majst01 (approved):

PR #539, opened 2024-06-17
Initial generic rethinkdb store

majst01:
Just to see how this would feel

### DIFF
--- a/cmd/metal-api/internal/datastore/rethinkdb.go
+++ b/cmd/metal-api/internal/datastore/rethinkdb.go
@@ -61,7 +61,7 @@ type RethinkStore struct {
 }
 
 // New creates a new rethink store.
-func New(log *slog.Logger, dbhost string, dbname string, dbuser string, dbpass string) *RethinkStore {
+func NewRethinkStore(log *slog.Logger, dbhost string, dbname string, dbuser string, dbpass string) *RethinkStore {
 	return &RethinkStore{
 		log:    log,
 		dbhost: dbhost,
@@ -344,6 +344,14 @@ tryAgain:
 		goto tryAgain
 	}
 	return s
+}
+
+func (rs *RethinkStore) DBName() string {
+	return rs.dbname
+}
+
+func (rs *RethinkStore) QueryExecutor() r.QueryExecutor {
+	return rs.session
 }
 
 func (rs *RethinkStore) findEntityByID(table *r.Term, entity interface{}, id string) error {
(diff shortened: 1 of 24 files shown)


ulrichSchreiner:
You use `metal.Entity` as a constraint of your type parameter. But the whole store only works, if the type is a pointer receiver for your constraint; so you always have to use pointer-types as type parameters: `*metal.IP`, `*metal.Machine` , ...

This makes the API a little akward: Take the `Create` method:

~~~go
func (rs *rethinkStore[E]) Create(ctx context.Context, e E) error
~~~

If you only look at the signature you do not see that the last parameter has to be a pointer to a type. Imho it is also not a very good style to have such an **output parameter**. Why not have a signature like

~~~go
func (rs *rethinkStore[E]) Create(ctx context.Context) (*E, error)
~~~

In this case it is clear that you return a pointer to your type parameter. But this requires your type constraint to be more tricky: I think you need two type parameters. One for the type without a constraint and the second with our `metal.Entity` interface which should contain also contain a type parameter which is a simple a pointer-receiver to the type parameter. so you can alwas create a pointer-receiver out of your type parameter to call `Set...`  methods (which only work on pointer receivers).

The [type parameters proposal](https://go.googlesource.com/proposal/+/HEAD/design/43651-type-parameters.md#pointer-method-example) has a good example for this problem.

cmd/metal-api/internal/metal/metal.go
```diff
@@ -97,6 +97,8 @@ type Entity interface {
 	GetCreated() time.Time
 	// SetCreated sets the entity's creation time
 	SetCreated(created time.Time)
+
+	TableName() string
```
ulrichSchreiner, line 101: Are you sure you want the `TableName` as a method in the entity interface? You are mixing regular values of the entity (id, created, ...) with a value which is only needed for the concrete storage.

majst01: good point, removed

cmd/metal-api/internal/datastore/rethinkdb.go
```diff
@@ -346,6 +346,14 @@ tryAgain:
 	return s
 }
 
+func (rs *RethinkStore) DBName() string {
+	return rs.dbname
```
majst01, line 350: not required

majst01: no

PR #551, opened 2024-07-22
Do not error out if no ip is available in a prefix

majst01:
closes: #550 
Depends on: https://github.com/metal-stack/go-ipam/pull/157


Gerrit91 (approved):

cmd/metal-api/internal/service/ip-service.go
```diff
@@ -439,8 +437,11 @@ func allocateSpecificIP(ctx context.Context, parent *metal.Network, specificIP s
 func allocateRandomIP(ctx context.Context, parent *metal.Network, ipamer ipam.IPAMer) (ipAddress, parentPrefixCidr string, err error) {
 	for _, prefix := range parent.Prefixes {
 		ipAddress, err = ipamer.AllocateIP(ctx, prefix)
-		if err != nil && errors.Is(err, goipam.ErrNoIPAvailable) {
```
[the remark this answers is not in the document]
majst01: Now

cmd/metal-api/internal/service/network-service.go
```diff
@@ -560,6 +561,12 @@ func (r *networkResource) freeNetwork(request *restful.Request, response *restfu
 	for _, prefix := range nw.Prefixes {
 		err = r.ipamer.ReleaseChildPrefix(ctx, prefix)
 		if err != nil {
+			var connectErr *connect.Error
+			if errors.As(err, &connectErr) {
```
Gerrit91, line 565: Probably we cannot skip here as then we have orphaned prefixes after freeing the network.

cmd/metal-api/internal/service/ip-service.go
```diff
@@ -439,8 +437,11 @@ func allocateSpecificIP(ctx context.Context, parent *metal.Network, specificIP s
...
-			continue
+		var connectErr *connect.Error
```
majst01, line 440: check for err != nil is missing here

cmd/metal-api/internal/service/network-service.go
```diff
@@ -730,6 +743,12 @@ func (r *networkResource) deleteNetwork(request *restful.Request, response *rest
 	for _, p := range nw.Prefixes {
 		err := r.ipamer.DeletePrefix(ctx, p)
 		if err != nil {
+			var connectErr *connect.Error
```
Gerrit91, line 746: Same here.

PR #553, opened 2024-07-31
Adapt healthcheck to new metal-lib.

Gerrit91:

PR #554, opened 2024-08-02
chore(deps): bump metal-lib

vknabel:
Uses the newly introduced healthstatus package


Gerrit91:
Oh. You did this as well. :sweat_smile: 
I merged this with #553 already.

PR #557, opened 2024-08-06
Rewrite grpc wait server test to stabilize CI.

Gerrit91:
This is a pretty complex set of changes, mostly tests are affected though.

Integration tests that start the gRPC server now use a real NSQ instead of a fake one. This has the advantage that we can remove development code snippets from our production code and we also increase test coverage.


cmd/metal-api/internal/grpc/boot-service-wait_integration_test.go
```diff
@@ -5,343 +5,341 @@ package grpc
...
 		}
 	}
-	return false
-}
 
-func (t *test) shutdown() {
-	t.stopMachineInstances()
-	t.stopApiInstances()
+	assert.Equal(te.t, te.numberMachineInstances-te.numberAllocations, waiting)
+	assert.Len(te.t, ms, te.numberMachineInstances)
```
Gerrit91, line 164: If we are really ambitious, we could use the remaining clients which are waiting for testing an "outage scenario", i.e. shutting down the grpc-servers, check that the clients do not give up and reconnect after bringing the servers back up (and vice versa).

majst01: lets try this

majst01 (approved):

cmd/metal-api/internal/grpc/boot-service-wait_integration_test.go
Gerrit91: Done.

PR #562, opened 2024-08-14
Make additional announcable cidrs configurable per tenant super network

majst01:
Actually we have the Pod- and Service- Cidrs hard coded in metal-core and add them to the route maps per machine switch port during the creation of the frr.conf on a switch.

With this we put this as a optional configuration into the super tenant network called `additionalannouncablecidrs` which can contain a slice of cidrs which should be added to the route map of every machine created in his tenant network.

In https://github.com/metal-stack/metal-core/pull/127 we already filter out duplicate entries in the route maps to keep backwards compatibility.

currently running in Test

TODO:

- [x] add a migration to add this field with the current default from metal-core to every super tenant network
- [ ] https://github.com/metal-stack/metalctl/pull/260


cmd/metal-api/internal/service/switch-service.go
```diff
@@ -1026,7 +1044,7 @@ func findSwitchReferencedEntities(s *metal.Switch, ds *datastore.RethinkStore) (
 	return p, ips.ByProjectID(), m, ss, nil
 }
 
-func makeSwitchResponseList(ss metal.Switches, ds *datastore.RethinkStore) ([]*v1.SwitchResponse, error) {
+func (r *switchResource) makeSwitchResponseList(ss metal.Switches, ds *datastore.RethinkStore) ([]*v1.SwitchResponse, error) {
```
vknabel, line 1047: for `makeSwitchResponse` you dropped the datastore parameter. Is there a reason not to do the same for `makeSwitchResponseList`?

majst01: Ah thanks, one moment

majst01: fixed

vknabel (approved):

cmd/metal-api/internal/service/network-service.go
```diff
@@ -409,6 +417,23 @@ func (r *networkResource) createNetwork(request *restful.Request, response *rest
...
+			return nil, errors.New("additionalannouncablecidrs can only be set in a private super network")
+		}
+		for _, cidr := range additionalCidrs {
+			_, err := netip.ParsePrefix(cidr)
+			if err != nil {
+				return nil, fmt.Errorf("given cidr:%q in additionalannouncablecidrs is malformed:%w", cidr, err)
+			}
+			result = append(result, cidr)
+		}
+	}
```
Gerrit91, line 433: Indent level can be reduced:

```suggestion
	if len(additionalCidrs) == 0 {
	  return nil
	}
	
        if !privateSuper {
		return nil, errors.New("additionalannouncablecidrs can only be set in a private super network")
	}
	
	...
```

mwennrich (approved):

cmd/metal-api/internal/service/network-service.go
majst01: yes, done

cmd/metal-api/internal/metal/network.go
```diff
@@ -207,17 +207,18 @@ func (p *Prefix) equals(other *Prefix) bool {
...
+	PartitionID                string            `rethinkdb:"partitionid" json:"partitionid"`
+	ProjectID                  string            `rethinkdb:"projectid" json:"projectid"`
+	ParentNetworkID            string            `rethinkdb:"parentnetworkid" json:"parentnetworkid"`
+	Vrf                        uint              `rethinkdb:"vrf" json:"vrf"`
+	PrivateSuper               bool              `rethinkdb:"privatesuper" json:"privatesuper"`
+	Nat                        bool              `rethinkdb:"nat" json:"nat"`
+	Underlay                   bool              `rethinkdb:"underlay" json:"underlay"`
+	Shared                     bool              `rethinkdb:"shared" json:"shared"`
+	Labels                     map[string]string `rethinkdb:"labels" json:"labels"`
+	AdditionalAnnouncableCIDRs []string          `rethinkdb:"additionalannouncablecidrs" json:"additionalannouncablecidrs" description:"list of cidrs which are added to the route maps per tenant private network, these are typically pod- and service cidrs, can only be set in a supernetwork"`
```
Gerrit91, line 221:
```suggestion
	AdditionalAnnouncableCIDRs []string          `rethinkdb:"additionalannouncablecidrs" json:"additionalannouncablecidrs" description:"list of cidrs which are added to the route maps per tenant private network, these are typically pod- and service cidrs, can only be set when private super is set to true"`
```

cmd/metal-api/internal/service/network-service.go
```diff
@@ -409,6 +417,28 @@ func (r *networkResource) createNetwork(request *restful.Request, response *rest
 	r.send(request, response, http.StatusCreated, v1.NewNetworkResponse(nw, usage))
 }
 
+func validateAdditionalAnnouncableCIDRs(additionalCidrs []string, privateSuper bool) ([]string, error) {
```
Gerrit91, line 420: Maybe more like this?

```suggestion
func validateAdditionalAnnouncableCIDRs(additionalCidrs []string, privateSuper bool) error {
```

Gerrit91 (approved):

cmd/metal-api/internal/service/switch-service.go
```diff
@@ -894,16 +911,16 @@ func makeBGPFilter(m metal.Machine, vrf string, ips metal.IPsMap) (v1.BGPFilter,
...
 	} else {
-		filter, err = makeBGPFilterMachine(m, ips)
+		filter, err = r.makeBGPFilterMachine(m, ips)
 	}
 
 	return filter, err
 }
 
-func makeSwitchNics(s *metal.Switch, ips metal.IPsMap, machines metal.Machines) (v1.SwitchNics, error) {
+func (r *switchResource) makeSwitchNics(s *metal.Switch, ips metal.IPsMap, machines metal.Machines) (v1.SwitchNics, error) {
```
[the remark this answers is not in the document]
majst01: Also done

PR #563, opened 2024-08-16
Only fetch required ips per switch describe instead of all

majst01:
During switch describe, which is called from every metal-core every 20 sec, all ips are listed and later filtered by project.
This is a lot of useless data fetched.

This was also found during the dualstack implementation, test was not adopted yet.


cmd/metal-api/internal/datastore/ip.go
```diff
@@ -97,6 +97,22 @@ func (rs *RethinkStore) SearchIPs(q *IPSearchQuery, ips *metal.IPs) error {
...
+// according to https://stackoverflow.com/questions/44424377/rethinkdb-query-filter-against-multiple-values-for-single-key it should :-)
+func (rs *RethinkStore) FindIPsByProjects(projects []string) (metal.IPs, error) {
+	q := *rs.ipTable()
+	q = q.Filter(func(row r.Term) r.Term {
+		return row.Field("projectid").Contains(func(nw r.Term) r.Term {
+			return nw.Field("projectid").Contains(r.Expr(projects))
+		})
+	})
+
+	ips := make([]metal.IP, 0)
```
Gerrit91, line 111:
```suggestion
	ips := metal.IPs{}
```

PR #565, opened 2024-08-26
Allow additional auditing backend based on TimescaleDB

Gerrit91:
## Description

References:
- https://github.com/metal-stack/mini-lab/pull/217 
- https://github.com/metal-stack/metal-lib/pull/151
- https://github.com/metal-stack/metal-roles/pull/310


cmd/metal-api/main.go
```diff
@@ -287,11 +287,19 @@ func init() {
...
-	rootCmd.Flags().String("auditing-api-key", "secret", "api key for the auditing service")
-	rootCmd.Flags().String("auditing-index-prefix", "auditing", "auditing index prefix")
-	rootCmd.Flags().String("auditing-index-interval", "@daily", "auditing index creation interval, can be one of @hourly|@daily|@monthly")
-	rootCmd.Flags().Int64("auditing-keep", 14, "the amount of indexes to keep until cleanup")
+
+	rootCmd.Flags().String("auditing-meili-url", "http://localhost:7700", "url of the auditing service")
+	rootCmd.Flags().String("auditing-meili-api-key", "secret", "api key for the auditing service")
+	rootCmd.Flags().String("auditing-meili-index-prefix", "auditing", "auditing index prefix")
+	rootCmd.Flags().String("auditing-meili-index-interval", "@daily", "auditing index creation interval, can be one of @hourly|@daily|@monthly")
+	rootCmd.Flags().Int64("auditing-meili-keep", 14, "the amount of indexes to keep until cleanup")
```
majst01, line 295: I would prefer to remove the support for meilisearch completely instead of adding the complexity of supporting multiple auditing backends

Gerrit91: Maybe we will have introduce multiple backends though if FI-TS wants to have Splunk connected. But maybe also we want to introduce the buffer service as well, which we were discussing as well.

majst01: in my opinion, metal-api should only have one auditing implementation, if other parties are interested in other auditing backends, then this should be solved at a later layer.
This is why i would like to vote for a dedicated auditing backend which has plug-able backends

Gerrit91: I also think this should be done but it comes with quite a lot of work and maybe it's acceptable to introduce this beforehand and then switch to the dedicated auditing layer because in production we see too many issues with Meilisearch currently. :(

majst01: I could live with small steps towards the final solution as well

cmd/metal-api/main.go
```diff
@@ -757,7 +765,7 @@ func initRestServices(audit auditing.Auditing, withauth bool, ipmiSuperUser meta
 		releaseVersion = pointer.Pointer(viper.GetString("release-version"))
 	}
 
-	restful.DefaultContainer.Add(service.NewAudit(logger.WithGroup("audit-service"), audit))
+	restful.DefaultContainer.Add(service.NewAudit(logger.WithGroup("audit-service"), pointer.FirstOrZero(audit)))
```
vknabel, line 768: Is there there some documentation about this behavior of only using the first auditing backend as source for the search?

cmd/metal-api/internal/grpc/grpc-server.go
```diff
@@ -130,16 +130,19 @@ func Run(cfg *ServerConfig) error {
...
 		}
-		auditStreamInterceptor, err := auditing.StreamServerInterceptor(cfg.Auditing, log.WithGroup("auditing-grpc"), shouldAudit)
-		if err != nil {
-			return err
-		}
-		auditUnaryInterceptor, err := auditing.UnaryServerInterceptor(cfg.Auditing, log.WithGroup("auditing-grpc"), shouldAudit)
-		if err != nil {
-			return err
+
+		for _, backend := range cfg.Auditing {
```
vknabel, line 134: here we need an if statement to skip not enabled backends. See https://github.com/metal-stack/metal-roles/pull/310/files#diff-81777600d83d4794dafee8e817a4b696b9c1a05fe38798329235b2ec72537cd1R229

Gerrit91: The audit backends are not appended to the slice in case they are not enabled, so I think there is no need to skip here.

cmd/metal-api/main.go
Gerrit91: I added another flag `--auditing-search-backend` to let the user decide a specific backend.

cmd/metal-api/internal/grpc/grpc-server.go
vknabel: It's a good idea to not append disabled backends, though I think it's confusing to have an `Enabled` field in Go but not using it. Could be either remove enabled from the struct or implement the skip here?

cmd/metal-api/main.go
```diff
@@ -287,11 +287,20 @@ func init() {
 	rootCmd.Flags().StringP("masterdata-certkeypath", "", "", "the tls certificate key to talk to the masterdata-api")
 
 	rootCmd.Flags().Bool("auditing-enabled", false, "enable auditing")
-	rootCmd.Flags().String("auditing-url", "http://localhost:7700", "url of the auditing service")
-	rootCmd.Flags().String("auditing-api-key", "secret", "api key for the auditing service")
-	rootCmd.Flags().String("auditing-index-prefix", "auditing", "auditing index prefix")
-	rootCmd.Flags().String("auditing-index-interval", "@daily", "auditing index creation interval, can be one of @hourly|@daily|@monthly")
-	rootCmd.Flags().Int64("auditing-keep", 14, "the amount of indexes to keep until cleanup")
+	rootCmd.Flags().String("auditing-search-backend", "", "the auditing backend used as a source for search in the audit service. if explicitly specified the first one configured is picked given the following order of precedence: timescaledb,meilisearch")
```
vknabel, line 290:
```suggestion
	rootCmd.Flags().String("auditing-search-backend", "", "the auditing backend used as a source for search in the audit service. if not specified the first one configured is picked given the following order of precedence: timescaledb,meilisearch")
```

cmd/metal-api/main.go
```diff
@@ -908,33 +916,74 @@ func evaluateVPNConnected() error {
...
+		backends = append(backends, backend)
+
+		if viper.GetString("auditing-search-backend") == "meilisearch" {
+			searchBackend = backend
+		}
+	}
+
+	if searchBackend == nil {
+		searchBackend = pointer.FirstOrZero(backends)
+	}
```
vknabel, line 974: If all backends above are nil, this will also be nil

```suggestion
```

cmd/metal-api/main.go
```diff
@@ -908,33 +916,74 @@ func evaluateVPNConnected() error {
...
+			Retention: viper.GetString("auditing-timescaledb-retention"),
+		})
+
+		if err != nil {
+			return nil, nil, err
+		}
+
+		backends = append(backends, backend)
+
+		if viper.GetString("auditing-search-backend") == "timescaledb" {
```
vknabel, line 947: can we... introduce constants for `timescaledb` and `meilisearch`? :innocent:

Gerrit91: Done.

vknabel (approved):

mwennrich (approved):

PR #566, opened 2024-09-04
Add switch migrate endpoint

iljarotar:
Replaces https://github.com/metal-stack/metal-api/pull/526

With the [previous PR](https://github.com/metal-stack/metal-api/pull/526) migrating from Cumulus to SONiC was expected to occur during switch registration. The problem with that is that a switch replacement requires the new switch to have the same ID as the old one. But deploying a new switch with the same ID as an existing one can be problematic in some cases. 

Therefore the new approach implements a `switch/migrate` endpoint that migrates the configuration from one switch to another, allowing both switches to have different OSs (Cumulus and Edgecore SONiC). With this approach it should be possible to register a switch first and perform the migration from an existing one to the new one afterwards, without requiring its ID to be the same as the old one's.


majst01:
It should be much easier to test with this approach in the mini-lab, simply create a setup there where 2 cumulus switches and one SONiC Switch is present and then migrate one of the cumulus switches to the SONiC Switch. Then the configuration on the SONiC Switch must be identical

iljarotar:
Open for discussion:

In addition to the migrate approach, a port mapping from Cumulus to SONiC also occurs when a switch gets replaced. So a migration is also possible right on switch registration. Do we want this, or should a switch replacement remain as it is, i.e. only for replacing a broken switch by a healthy one without changing OS?

iljarotar:
Turns out the mapping from Cumulus to SONiC port names is off by 1. `spw1` should map to `Ethernet0` etc.

cmd/metal-api/internal/datastore/switch.go
```diff
@@ -177,9 +177,21 @@ func (rs *RethinkStore) ConnectMachineWithSwitches(m *metal.Machine) error {
 	if err != nil {
 		return err
 	}
+	// e.g. "swp0s0" -> "Ethernet0"
+	dictionary, err := s1.MapPortNames(s2.OS.Vendor)
```
majst01, line 181: dictionary is a bit to generic, what about switchPortMapping ?

cmd/metal-api/internal/metal/network.go
```diff
@@ -303,9 +303,12 @@ func (nics Nics) FilterByHostname(hostname string) (res Nics) {
 	return res
 }
 
+// NicMap maps nic names to the corresponding nics
+type NicMap map[string]*Nic
```
majst01, line 307: This is a useful refactoring, which could be done in a separate PR which will then reduce this PR

cmd/metal-api/internal/service/switch-service.go
```diff
@@ -562,6 +571,64 @@ func (r *switchResource) registerSwitch(request *restful.Request, response *rest
...
+	err = r.migrateMachineConnections(old, s)
+	if err != nil {
+		r.sendError(request, response, defaultError(err))
+		return
+	}
+
+	err = r.ds.UpdateSwitch(new, s)
+	if err != nil {
+		r.sendError(request, response, defaultError(err))
+		// TODO: should machine connections be rolled back in case of error
```
majst01, line 619: Error is then raised to the caller, and this error should contain instructions how to proceed. Either make a second call to migrate back or ??

cmd/metal-api/internal/service/v1/switch.go
```diff
@@ -32,9 +32,10 @@ type SwitchBase struct {
 }
 
 type SwitchOS struct {
-	Vendor           string `json:"vendor" description:"the operating system vendor the switch currently has" optional:"true"`
-	Version          string `json:"version" description:"the operating system version the switch currently has" optional:"true"`
-	MetalCoreVersion string `json:"metal_core_version" description:"the version of metal-core running" optional:"true"`
+	// TODO: do we need a distribution e.g. edgecore vs broadcom?
```
majst01, line 35: I would remove this comment, actually we do not need it.

majst01 (approved):

cmd/metal-api/internal/service/v1/switch.go
```diff
@@ -32,9 +32,9 @@ type SwitchBase struct {
 }
 
 type SwitchOS struct {
-	Vendor           string `json:"vendor" description:"the operating system vendor the switch currently has" optional:"true"`
-	Version          string `json:"version" description:"the operating system version the switch currently has" optional:"true"`
-	MetalCoreVersion string `json:"metal_core_version" description:"the version of metal-core running" optional:"true"`
+	Vendor           metal.SwitchOSVendor `json:"vendor" description:"the operating system vendor the switch currently has" optional:"true"`
```
majst01, line 35:
```suggestion
	Vendor           metal.SwitchOSVendor `json:"vendor" description:"the operating system vendor the switch currently has" optional:"true" enum:"SONiC|Cumulus"`
```

cmd/metal-api/internal/metal/switch.go
```diff
@@ -144,3 +155,203 @@ func (s *Switch) SetVrfOfMachine(m *Machine, vrf string) {
...
+}
+
+func portNameToLine(port string, os SwitchOSVendor) (int, error) {
+	if os == SwitchOSVendorSonic {
+		return sonicPortNameToLine(port)
+	}
+	if os == SwitchOSVendorCumulus {
+		return cumulusPortNameToLine(port)
+	}
+	return 0, fmt.Errorf("unknow switch os %s", os)
```
majst01, line 266:
```suggestion
	return 0, fmt.Errorf("unknown switch os %s", os)
```

cmd/metal-api/internal/metal/switch.go
```diff
@@ -24,11 +26,20 @@ type Switch struct {
...
 
+// SwitchOSVendor is an enum denoting the name of a switch OS
+type SwitchOSVendor string
+
+// The enums for switch OS vendors
+const (
+	SwitchOSVendorSonic   SwitchOSVendor = "SONiC"
+	SwitchOSVendorCumulus SwitchOSVendor = "Cumulus"
+)
+
```
majst01, line 42: please add a 
```go
func ValidateSwitchOSVendor(os SwitchOSVendor) error {

}
```

and call it in the switch-service.go#register|migrate

cmd/metal-api/internal/service/switch-service.go
```diff
@@ -562,6 +571,71 @@ func (r *switchResource) registerSwitch(request *restful.Request, response *rest
...
+
+	new, err := r.ds.FindSwitch(requestPayload.NewSwitchID)
+	if err != nil {
+		r.sendError(request, response, defaultError(err))
+		return
+	}
+
+	if old.RackID != new.RackID {
+		r.sendError(request, response, httperrors.BadRequest(fmt.Errorf("new switch must be in the same rack as the old one")))
+	}
```
majst01, line 606: return missing

cmd/metal-api/internal/service/switch-service.go
```diff
@@ -562,6 +571,71 @@ func (r *switchResource) registerSwitch(request *restful.Request, response *rest
...
+
+	if len(new.MachineConnections) > 0 {
+		r.sendError(request, response, httperrors.BadRequest(fmt.Errorf("target switch already has machine connections")))
+	}
```
majst01, line 610: return missing

majst01 (changes requested):

majst01 (approved):

majst01:
We should wait for a retest in the fleet.

cmd/metal-api/internal/metal/switch.go
```diff
@@ -81,6 +92,13 @@ func SwitchModeFrom(name string) SwitchMode {
 	}
 }
 
+func ValidateSwitchOSVendor(os SwitchOSVendor) error {
+	if os != SwitchOSVendorCumulus && os != SwitchOSVendorSonic {
+		return fmt.Errorf("unknown switch os vendor %s", os)
+	}
+	return nil
```
Gerrit91, line 99: This is a bit easier to maintain because the linter figures out if it's exhaustive or not:

```suggestion
	switch os {
	case SwitchOSVendorCumulus, SwitchOSVendorSonic:
		return nil
	default:
		return fmt.Errorf("unknown switch os vendor %s", os)
	}
```

cmd/metal-api/internal/metal/switch.go
```diff
@@ -144,3 +162,203 @@ func (s *Switch) SetVrfOfMachine(m *Machine, vrf string) {
...
+	}
+
+	if targetOS == SwitchOSVendorCumulus {
+		return cumulusPortByLineNumber(line, allLines), nil
+	}
+	if targetOS == SwitchOSVendorSonic {
+		return sonicPortByLineNumber(line), nil
+	}
+
+	return "", fmt.Errorf("unknown target switch os %s", targetOS)
```
Gerrit91, line 249: Prefer switch statement here to easier guarantee exhaustiveness as mentioned above.

cmd/metal-api/internal/metal/switch.go
```diff
@@ -144,3 +162,203 @@ func (s *Switch) SetVrfOfMachine(m *Machine, vrf string) {
...
+}
+
+func portNameToLine(port string, os SwitchOSVendor) (int, error) {
+	if os == SwitchOSVendorSonic {
+		return sonicPortNameToLine(port)
+	}
+	if os == SwitchOSVendorCumulus {
+		return cumulusPortNameToLine(port)
+	}
+	return 0, fmt.Errorf("unknown switch os %s", os)
```
Gerrit91, line 273: Prefer switch statement here to easier guarantee exhaustiveness as mentioned above.

cmd/metal-api/internal/service/switch-service.go
```diff
@@ -132,6 +132,15 @@ func (r *switchResource) webService() *restful.WebService {
 		Returns(http.StatusOK, "OK", v1.SwitchNotifyResponse{}).
 		DefaultReturns("Error", httperrors.HTTPErrorResponse{}))
 
+	ws.Route(ws.POST("/migrate").
+		To(editor(r.migrate)).
```
Gerrit91, line 136: I think we should raise this to admin, which is typically an operator. Components like metal-core are editor.

```suggestion
		To(admin(r.migrate)).
```

Gerrit91 (commented):
Just comments on style. I guess the technical aspects are spot on.

Gerrit91 (approved):

PR #570, opened 2024-09-11
Configure DNS servers in machine allocation optionally

simcod:


cmd/metal-api/internal/service/v1/machine.go
```diff
@@ -45,6 +45,7 @@ type MachineAllocation struct {
 	VPN              *MachineVPN               `json:"vpn" description:"vpn connection info for machine" optional:"true"`
 	AllocationUUID   string                    `json:"allocationuuid" description:"a unique identifier for this machine allocation, can be used to distinguish between machine allocations over time."`
 	FirewallRules    *FirewallRules            `json:"firewall_rules,omitempty" description:"a set of firewall rules to apply" optional:"true"`
+	DNSServers       []string                  `json:"dns_servers,omitempty" description:"the dns servers used for the machine" optional:"true"`
```
majst01, line 48: you must also put the DNSService into the response

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -1237,6 +1239,7 @@ func allocateMachine(ctx context.Context, logger *slog.Logger, ds *datastore.Ret
 		VPN:             allocationSpec.VPN,
 		FirewallRules:   firewallRules,
 		UUID:            uuid.New().String(),
+		DNSServers:      allocationSpec.DNSServers,
```
majst01, line 1242: some sort of validation is required like:
walk through all elements
if the element is a ip, validate with netip.ParseAddr()
else it must be a rfc conformant hostname

majst01: This ensures a breaking OS installation

simcod:
Will be handled in a PR with dns and ntp combined.

PR #571, opened 2024-09-12
Configure DNS and NTP in machine allocation

simcod:
This PR is related to [MEP-14](https://docs.metal-stack.io/dev/development/proposals/MEP14/README/).

Related PRs:

- [ ] https://github.com/metal-stack/metal-hammer/pull/139
- [ ] https://github.com/metal-stack/metal-images/pull/255
- [ ] https://github.com/metal-stack/metalctl/pull/266
- [ ] https://github.com/metal-stack/docs/pull/222


robertvolkmann:
"It should be a feature of `metal-stack` and not reliant on Gardener-specific features like `CloudProfile`. For one customer, having the ability to configure it directly within the `Partition` would be advantageous. Additionally, for isolated clusters, the DNS servers and NTP servers in the `CloudProfile` would overwrite the values provided in the `Partition`. I also tend to require configured DNS and NTP servers for every partition in the future to eliminate any default values in the metal images."

majst01:
> It should be a feature of `metal-stack` and not reliant on Gardener-specific features like `CloudProfile`. For one customer, having the ability to configure it directly within the `Partition` would be advantageous. Additionally, for isolated clusters, the DNS servers and NTP servers in the `CloudProfile` would overwrite the values provided in the `Partition`. I also tend to require configured DNS and NTP servers for every partition in the future to eliminate any default values in the metal images.

On the other hand, properties like `image` or `firewall-image` are not defaulted and must be provided during allocation by the GEPM. 

For me it is a bit opinionated if we inherit from the partition.

robertvolkmann:
How do you plan to provide the NTP configuration for `metal-hammer` if the NTP servers cannot be configured in the `Partition`?

majst01:
> How do you plan to provide the NTP configuration for `metal-hammer` if the NTP servers cannot be configured in the `Partition`?

This is actually done in the pixie-core deployment, but would also be easier if partition contains these configuration.

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -1143,6 +1147,32 @@ func createMachineAllocationSpec(ds *datastore.RethinkStore, machineRequest v1.M
...
+		return nil, errors.New("please specify a maximum of three dns servers")
+	}
+	for _, dnsip := range machineRequest.DNSServers {
+		_, err := netip.ParseAddr(dnsip.IP)
+		if err != nil {
+			return nil, fmt.Errorf("IP: %s for DNS server not correct err: %w", dnsip, err)
+		}
+	}
+
+	if len(machineRequest.NTPServers) <= 3 || len(machineRequest.NTPServers) > 5 {
```
Gerrit91, line 1160: Should an empty list be allowed, too? In the MEP it says that specifying the NTP servers is optional. And if not provided the distro defaults are used.

majst01: Sure, can be totally omitted, but if given len must be between 3 and 5

@simcod this must be fixed obviously

Gerrit91:
Regarding the defaulting through the partition entity: I am actually open to offer this defaulting layer when for certain environments this makes everything easier to configure. It seems there can be partitions that do not reach the internet at all and no machine can be provisioned without passing custom DNS and NTP servers. For these scenarios it's really cumbersome to always pass these settings all the time. As long as it's optional to provide, it does not hurt to add this.

majst01:
@simcod please add matching fields to the Partition entity (v1 and database) and check in the service if they are set and use the partition config for defaults.

cmd/metal-api/internal/service/v1/partition.go
majst01, line 1: please add DNSServers and NTPServer to `NewPartitionResponse` func here as well

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -1143,6 +1147,51 @@ func createMachineAllocationSpec(ds *datastore.RethinkStore, machineRequest v1.M
...
+		dnsServers = machineRequest.DNSServers
+	} else {
+		dnsServers = partition.DNSServers
+	}
+	for _, dnsip := range dnsServers {
+		_, err := netip.ParseAddr(dnsip.IP)
+		if err != nil {
+			return nil, fmt.Errorf("IP: %s for DNS server not correct err: %w", dnsip, err)
+		}
+	}
```
majst01, line 1171:
```suggestion
	for _, dnsserver := range dnsServers {
		_, err := netip.ParseAddr(dnsserver.IP)
		if err != nil {
			return nil, fmt.Errorf("ip: %s for dns server not correct err: %w", dnsserver, err)
		}
	}
```

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -1143,6 +1147,51 @@ func createMachineAllocationSpec(ds *datastore.RethinkStore, machineRequest v1.M
...
+		ntpServers = machineRequest.NTPServers
+	} else {
+		ntpServers = partition.NTPServers
+	}
+
+	for _, ntpserver := range ntpServers {
+		if net.ParseIP(ntpserver.Address) != nil {
+			_, err := netip.ParseAddr(ntpserver.Address)
+			if err != nil {
+				return nil, fmt.Errorf("IP: %s for NTP server not correct err: %w", ntpserver, err)
```
majst01, line 1186:
```suggestion
				return nil, fmt.Errorf("ip: %s for ntp server not correct err: %w", ntpserver, err)
```

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -1143,6 +1147,51 @@ func createMachineAllocationSpec(ds *datastore.RethinkStore, machineRequest v1.M
...
+			}
+		} else {
+			if !govalidator.IsDNSName(ntpserver.Address) {
+				return nil, fmt.Errorf("DNS name: %s for NTP server not correct err: %w", ntpserver, err)
+			}
```
majst01, line 1191: - Is only checked if ntpserver.Address == nil
- DNS and NTP should be lowercase

cmd/metal-api/internal/metal/machine.go
```diff
@@ -175,6 +177,18 @@ type IngressRule struct {
 	Comment  string   `rethinkdb:"comment" json:"comment"`
 }
 
+type DNSServers []DNSServer
+
+type DNSServer struct {
+	IP string `rethinkdb:"ip" json:"ip"`
```
majst01, line 183:
```suggestion
	IP string `rethinkdb:"ip" json:"ip" description:"the ip address of this dns server"`
```

cmd/metal-api/internal/metal/machine.go
```diff
@@ -175,6 +177,18 @@ type IngressRule struct {
...
+}
+
+type NTPServers []NTPServer
+
+type NTPServer struct {
+	Address string `address:"address" json:"address"`
```
majst01, line 189:
```suggestion
	Address string `address:"address" json:"address" description:"ip address or dns hostname of this ntpserver"`
```

majst01 (changes requested):

cmd/metal-api/internal/service/machine-service.go
majst01: my bad

majst01 (approved):

cmd/metal-api/internal/service/machine-service.go
```diff
@@ -1143,6 +1147,51 @@ func createMachineAllocationSpec(ds *datastore.RethinkStore, machineRequest v1.M
 		return nil, fmt.Errorf("size:%s not found err:%w", sizeID, err)
 	}
 
+	partition, err := ds.FindPartition(partitionID)
+	if err != nil {
+		return nil, fmt.Errorf("partition:%s not found err:%w", partitionID, err)
+	}
+	var (
+		dnsServers metal.DNSServers
+		ntpServers metal.NTPServers
```
Gerrit91, line 1156: This would render the `else` branches unnecessary:

```suggestion
		dnsServers = partition.DNSServers
		ntpServers = partition.NTPServers
```

cmd/metal-api/internal/service/partition-service.go
```diff
@@ -205,6 +205,16 @@ func (r *partitionResource) createPartition(request *restful.Request, response *
...
 
+	var dnsServers metal.DNSServers
+	if len(requestPayload.DNSServers) > 0 {
+		dnsServers = requestPayload.DNSServers
+	}
+
+	var ntpServers metal.NTPServers
+	if len(requestPayload.NTPServers) > 0 {
+		ntpServers = requestPayload.NTPServers
+	}
```
Gerrit91, line 216: Same validations as in the machine spec should be implemented here. Otherwise machines might default to the partition defaults, which are invalid.

cmd/metal-api/internal/service/partition-service.go
```diff
@@ -206,13 +206,23 @@ func (r *partitionResource) createPartition(request *restful.Request, response *
 	}
 
 	var dnsServers metal.DNSServers
-	if len(requestPayload.DNSServers) > 0 {
-		dnsServers = requestPayload.DNSServers
+	reqDNSServers := requestPayload.DNSServers
+	if len(reqDNSServers) > 0 {
```
majst01, line 210: i think this check can be omited, because the validate func iterates over the dnsservers otherwise returns nil

cmd/metal-api/internal/service/partition-service.go
```diff
@@ -206,13 +206,23 @@ func (r *partitionResource) createPartition(request *restful.Request, response *
...
+		}
+
+		dnsServers = reqDNSServers
 	}
 
 	var ntpServers metal.NTPServers
-	if len(requestPayload.NTPServers) > 0 {
-		ntpServers = requestPayload.NTPServers
+	reqNtpServers := requestPayload.NTPServers
+	if len(ntpServers) > 0 {
```
majst01, line 221: same here

Gerrit91 (approved):

PR #580, opened 2024-10-04
Add powersupply failures to machine issues

majst01:
Now we have the information about the power supply state, evaluate this info and report as machine issue if something is wrong here.


cmd/metal-api/internal/issues/powersupply-failure.go
```diff
@@ -0,0 +1,40 @@
...
+		Type:        TypePowerSupplyFailure,
+		Severity:    SeverityMajor,
+		Description: "machine has power supply failures",
+		RefURL:      "https://docs.metal-stack.io/stable/installation/troubleshoot/#power-supply-failure",
+	}
+}
+
+func (i *issuePowerSupplyFailure) Evaluate(m metal.Machine, ec metal.ProvisioningEventContainer, c *Config) bool {
+	for _, ps := range m.IPMI.PowerSupplies {
+		if ps.Status.Health != "OK" || ps.Status.State != "Enabled" {
```
majst01, line 30: Would be nice if these String values are somewhere available and documented

majst01: maybe we should at least compare case insensitive

cmd/metal-api/internal/issues/issues_test.go
```diff
@@ -85,6 +85,39 @@ func TestFindIssues(t *testing.T) {
 				}
 			},
 		},
+		{
+			name: "powersupply failure",
+			only: []Type{TypePowerSupplyFailure},
+			machines: func() metal.Machines {
+				noPartitionMachine := machineTemplate("power-supply-failure")
```
Gerrit91, line 92:
```suggestion
				defectPowerSupplyMachine := machineTemplate("power-supply-failure")
```

PR #582, opened 2024-10-10
Configurable interface and neighbor count for machines

majst01:
## References

Closes None.

## Additional Description

None

## Release Notes

Configurable Interfaces and Neighbors for machines which are required to put them into waiting state.


majst01:
Will not be merged

PR #598, opened 2024-11-25
feat: add support for reporting of switch BGP states by metal-core

### LINKED ISSUE #590: Detect if a machine has both uplinks up
In Order to prepare the Cumulus SONiC migration we must ensure that every machine is connected physically and logically to both switches.

We could achieve this by using the lldp messages every machine sends to both switches, and the metal-core reports them back to metal-api. But actually we only calculate the `Phoned Home` state, regardless if one or two switches report that state.

This should be changed that:

- [ ] Phoned Home is considered as such if at least one switch is reporting as before
- [ ] Add additional infos to the MachineConnection, which connection reports LLDP messages and when 
- [ ] enhance `metalctl` to visualize machines which have a broken uplink

mwennrich:
## Description

add support for reporting of switch BGP states by metal-core

Closes #590

References:
- https://github.com/metal-stack/metal-core/pull/136
- https://github.com/metal-stack/metal-roles/pull/352

### DIFF
--- a/cmd/metal-api/internal/service/switch-service.go
+++ b/cmd/metal-api/internal/service/switch-service.go
@@ -6,6 +6,7 @@ import (
 	"log/slog"
 	"net/http"
 	"net/netip"
+	"reflect"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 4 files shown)


cmd/metal-api/internal/service/switch-service.go
```diff
@@ -325,6 +326,35 @@ func (r *switchResource) notifySwitch(request *restful.Request, response *restfu
...
+		for i, nic := range newSwitch.Nics {
+			bpsnew, ok := requestPayload.BGPPortStates[nic.Name]
+			if !ok && nic.BGPPortState == nil {
+				continue
+			}
+			if nic.BGPPortState == nil {
+				newSwitch.Nics[i].BGPPortState = nil
+				switchUpdated = true
+			}
+			if reflect.DeepEqual(nic.BGPPortState, &bpsnew) {
```
majst01, line 340: can we detect equality by comparing members instead ? Using reflect.DeepEqual is not really recommended in production code

majst01 (commented):
Nice, should be enhanced later to have reports by machine

majst01 (approved):

PR #604, opened 2025-01-24
Update to headscale v0.26.1

majst01:
## Description

Update to headscale v0.26.1

TODO:
- [ ] update release vector to update the headscale deployment as well.
- [ ] Check which tailscale versions are connected in production, as older versions than v1.62 are no longer supported


majst01:
closing in favour of #649

PR #605, opened 2025-01-30
If the grpc certificate changed, reboot metal-hammer

majst01:
## Description

During the tests to rotate the ca and all certificates, waiting metal-hammer instances did not reboot. 
Fix this in a ugly manner.

Related: https://github.com/metal-stack/metal-hammer/issues/150

### DIFF
--- a/pkg/grpc/wait.go
+++ b/pkg/grpc/wait.go
@@ -2,11 +2,11 @@ package helper
 
 import (
 	"context"
-	"crypto/x509"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


mwennrich (approved):

Gerrit91:
Theoretically I think we could try it with `errors.Is`, too: https://github.com/golang/go/blob/e81f7155154c0f5d40363e84a8f24a5b559b5eed/src/crypto/x509/verify.go#L141-L149

PR #549, opened 2024-07-17
Dualstack Network Support

majst01:
Closes #164 

## Actions Required

```ACTIONS_REQUIRED
The ChildPrefixLength property was removed from the Partition Entity. Every TenantSupernetwork must now be configured with ChildPrefixLength accordingly.
This is done during startup in the database migration step and set to the same value as in the `Partition`.

It is required to add a defaultchildprefixlength property to the tenant super network in you configuration because otherwise the initial network creation will fail:
Sample:
  defaultchildprefixlength:
    IPv4: 22

Also ensure that all firewall-controllers are up-to-date in all firewalls before adding ipv6 prefixes to any of the exiting networks.
``` 

Depends on:

- [ ] https://github.com/metal-stack/docs/pull/202
- [x] https://github.com/metal-stack/metalctl/pull/256
- [x] https://github.com/metal-stack/mini-lab/pull/184
- [x] https://github.com/metal-stack/metal-core/pull/127
- [x] https://github.com/metal-stack/metal-networker/pull/110
- [x] https://github.com/metal-stack/metal-images/pull/252
- [ ] https://github.com/metal-stack/gardener-extension-provider-metal/issues/439
- [x] https://github.com/metal-stack/metal-networker/pull/114
- [x] https://github.com/metal-stack/metal-networker/issues/119

replaces #544 

TODO:

- [ ] add possibility to create machines in different networks like internet and tenant-network, but only from tenant-network IPs from both AF are acquired, but only ipv4 is possible from internet even if dualstack

Issue #16, opened 2020-03-13
Allow acquiring child prefixes from other networks than private super networks


majst01:
We simplified managing firewall rules for FITS Classic networking for our tenants by splitting the MPLS network manually into pieces:

```
  mpls-nbg-w8101-dkb                       MPLS Network for nbg-w8101 DKB                                        nbg-w8101  true   100.127.130.0/24     ●   
  mpls-nbg-w8101-dwp                       MPLS Network for nbg-w8101 DWP                                        nbg-w8101  true   100.127.131.0/24     ●   
  mpls-nbg-w8101-fits                      MPLS Network for nbg-w8101                                            nbg-w8101  true   100.127.129.0/24     ●   
                                           FI-TS                                                                                                              
  mpls-nbg-w8101-nlb-dev                   MPLS Network for nbg-w8101                                            nbg-w8101  true   100.127.132.0/28     ●   
                                           NordLB Dev                                                                                                         
```

This is quite a static approach and it is to be expected that this will become hard to manage for us. For example, we have to prevent that tenant A can start firewalls in the tenant B network. But how can we decide this when there are no projects related to these networks? (only static config mappings come to my mind to achieve this)

The idea is to use the existing IPAM child prefix acquisition to enable users to dynamically carve out subnetworks of the MPLS network. These subnetworks would be smaller and belong to a project, therefore we could make a decision if acquiring IPs in this network is allowed or not.

majst01:
Additional goal: "Wish prefix" -> Let a user try to acquire a certain prefix
(would be also very good for Gardener in case something goes wrong there)

PR #609, opened 2025-02-19
Remove unnecessary parent network ID field from network allocation.

Gerrit91:
## Description

I think this is unnecessary because so far a super network never has a parent? I think this field would imply nested super networks? Was added in #549.


Gerrit91:
I think it's not possible to create a super network that has a parent? This feature is probably not thought through and would implement https://github.com/metal-stack/metal-api/issues/16.

majst01 (approved):

PR #612, opened 2025-03-17
Update semver parser with breaking change regarding image versions

majst01:
## Description

Mastermind/semv v3.3.1 fix a longstanding bug regarding invalid semantic versions. Leading zeros have been valid before, now its an invalid version.

Fix this.


mwennrich (approved):

Gerrit91:
I think this was a breaking change, it for example broke the mini-lab. This requires a minor version change including a breaking change section for the release notes.

PR #615, opened 2025-03-25
Add check to prevent migrating a machine from one rack to another wit…

### LINKED ISSUE #613: Prevent recabling machine to a different rack in a partition
When recabling a machine from one rack to another, the system ends up into a state where we have four switches holding the machine connections.

This is due to a implementation detail in the machine registration where we do not have the check for validating that a machine connection already exists in the database for other switches in the partition.

We should add this additional check in order to enforce machine deletion from database when a machine is recabled.

iljarotar:
…hout deleting it first

## Description

Closes #613.


Gerrit91 (commented):
To me it looks good, I will try it out in a test environment.

Gerrit91 (approved):

PR #620, opened 2025-05-16
Provide network type migration as preparation for MEP-4.

Gerrit91:
## Description

References https://github.com/metal-stack/metal-apiserver/pull/14.

### DIFF
--- a/cmd/metal-api/internal/datastore/migrations/09_networktype.go
+++ b/cmd/metal-api/internal/datastore/migrations/09_networktype.go
@@ -1,0 +1,96 @@
+package migrations
+
+import (
+	"slices"
+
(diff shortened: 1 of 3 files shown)


majst01:
During implementation of machine-service in the metal-apiserver it turns out that we also need to migrate the machine.Allocation.MachineNetworks as well. Otherwise this expensive translation must be done during every machine get/list call.

majst01:
> During implementation of machine-service in the metal-apiserver it turns out that we also need to migrate the machine.Allocation.MachineNetworks as well. Otherwise this expensive translation must be done during every machine get/list call.

I implemented the lookup in the get request, no more migration required.

PR #625, opened 2025-07-17
Spread only on same machine size.

Gerrit91:
## Description

Looks like currently the machine spread is calculated considering all machine sizes, which can cause unwanted distribution.

### DIFF
--- a/cmd/metal-api/internal/datastore/machine_integration_test.go
+++ b/cmd/metal-api/internal/datastore/machine_integration_test.go
@@ -1073,9 +1073,7 @@ func Test_FindWaitingMachine_RackSpreadingDistribution(t *testing.T) {
 	defer func() {
 		_, err := sharedDS.machineTable().Delete().RunWrite(sharedDS.session)
 		require.NoError(t, err)
-	}()
-	defer func() {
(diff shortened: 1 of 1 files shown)


cmd/metal-api/internal/datastore/machine_integration_test.go
```diff
@@ -1061,3 +1059,123 @@ func Test_FindWaitingMachine_NoConcurrentModificationErrors(t *testing.T) {
...
+	)
+
+	defer func() {
+		_, err := sharedDS.machineTable().Delete().RunWrite(sharedDS.session)
+		require.NoError(t, err)
+	}()
+	defer func() {
+		_, err := sharedDS.eventTable().Delete().RunWrite(sharedDS.session)
+		require.NoError(t, err)
+	}()
```
majst01, line 1080:
```suggestion
	defer func() {
		_, err := sharedDS.machineTable().Delete().RunWrite(sharedDS.session)
		require.NoError(t, err)
		_, err = sharedDS.eventTable().Delete().RunWrite(sharedDS.session)
		require.NoError(t, err)
	}()
```

PR #632, opened 2025-09-02
Consider all machines of partition for evaluating faulty machines

### LINKED ISSUE #623: Consider all machines for evaluating faulty machines in partition capacity endpoint
Otherwise we get some funny behavior with filters:

```
❯ m partition capacity --id partition-a --size m1-large-x86
PARTITION    SIZE           ALLOCATED   FREE   UNAVAILABLE   RESERVATIONS   |   TOTAL   |   FAULTY 
partition-a  m1-large-x86   20          2      0             2 (3/5 used)   |   24      |   0       

❯ m partition capacity | grep partition-a | grep m1-large-x86
partition-a  m1-large-x86   20          2      0             2 (3/5 used)     |   24      |   8
```

ostempel:
## Description

Closes #623 

Consider all machines of partition (if partition filter is set) for evaluating faulty machines.


cmd/metal-api/internal/service/partition-service.go
```diff
@@ -413,6 +414,16 @@ func (r *partitionResource) calcPartitionCapacity(pcr *v1.PartitionCapacityReque
...
 	}
 
+	// if filtered on partition get all without more filters for issues evaluation
+	if machineQuery.PartitionID != nil {
+		err = r.ds.SearchMachines(&datastore.MachineSearchQuery{PartitionID: machineQuery.PartitionID}, &allMs)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		allMs = ms
```
Gerrit91, line 424: `ms` could be filtered by size, so these are not all machines. There should be no other filters applied for machine evaluation (except maybe partition).

Gerrit91 (approved):
Thanks, hopefully it will fix the issue in production, too, but I think it will!

PR #635, opened 2025-12-02
implement image check for oci pull

mac641:
### References

- https://github.com/metal-stack/metal-hammer/pull/169
- https://github.com/metal-stack/metal-images/pull/375
- https://github.com/metal-stack/metal-hammer/pull/175
- https://github.com/metal-stack/pixie/pull/40

### DIFF
--- a/cmd/metal-api/internal/service/image-service.go
+++ b/cmd/metal-api/internal/service/image-service.go
@@ -352,26 +352,25 @@ func (r *imageResource) createImage(request *restful.Request, response *restful.
 }
 
 func checkImageURL(id, inputURL string, ociCredentials authn.Authenticator) error {
-	parsedURL := strings.Split(inputURL, "://")
+	parsedURL, err := url.Parse(inputURL)
+	if err != nil {
+		return fmt.Errorf("inputURL: %q could not be parsed, error: %w", inputURL, err)
+	}
 
-	switch parsedURL[0] {
+	switch parsedURL.Scheme {
 	case "http", "https":
-		_, err := url.ParseRequestURI(inputURL)
+		res, err := http.Head(parsedURL.String())
 		if err != nil {
-			return fmt.Errorf("image:%s could not be parsed. error:%w", inputURL, err)
-		}
-
-		res, err := http.Head(inputURL)
-		if err != nil {
-			return fmt.Errorf("image:%s is not accessible under:%s error:%w", id, inputURL, err)
+			return fmt.Errorf("image: %q is not accessible under:%s, error: %w", id, inputURL, err)
 		}
 		if res.StatusCode >= 400 {
-			return fmt.Errorf("image:%s is not accessible under:%s status:%s", id, inputURL, res.Status)
+			return fmt.Errorf("image:%q is not accessible under:%s, status:%s", id, inputURL, res.Status)
 		}
 	case "oci":
-		ref, err := name.ParseReference(parsedURL[1])
+		parsedURLWithoutScheme := strings.TrimPrefix(parsedURL.String(), "oci://")
+		ref, err := name.ParseReference(parsedURLWithoutScheme)
 		if err != nil {
-			return fmt.Errorf("image reference:%s could not be parsed. error:%w", inputURL, err)
+			return fmt.Errorf("image reference: %q could not be parsed, error: %w", inputURL, err)
 		}
 
 		if ociCredentials == nil {
@@ -380,10 +379,10 @@ func checkImageURL(id, inputURL string, ociCredentials authn.Authenticator) erro
 
 		_, err = remote.Image(ref, remote.WithAuth(ociCredentials))
 		if err != nil {
-			return fmt.Errorf("image:%s is not accessible under:%s error:%w", id, inputURL, err)
+			return fmt.Errorf("image: %q is not accessible under: %s, error: %w", id, inputURL, err)
 		}
 	default:
-		return fmt.Errorf("image:%s with url:%s has unkown protocol", id, inputURL)
+		return fmt.Errorf("image: %q with url: %s has unknown protocol", id, inputURL)
 	}
 
 	return nil
--- a/cmd/metal-api/internal/service/image-service_test.go
+++ b/cmd/metal-api/internal/service/image-service_test.go
@@ -111,7 +111,7 @@ func TestDeleteImage(t *testing.T) {
 func TestCheckImageURL(t *testing.T) {
 	t.Run("Invalid URL", func(t *testing.T) {
 		err := checkImageURL("testID", "http://invalid url", nil)
-		require.EqualError(t, err, "image:http://invalid url could not be parsed. error:parse \"http://invalid url\": invalid character \" \" in host name")
+		require.EqualError(t, err, "inputURL: \"http://invalid url\" could not be parsed, error: parse \"http://invalid url\": invalid character \" \" in host name")
 	})
 
 	t.Run("HTTP URL with successful HEAD request", func(t *testing.T) {
@@ -131,12 +131,12 @@ func TestCheckImageURL(t *testing.T) {
 		defer server.Close()
 
 		err := checkImageURL("testID", server.URL, nil)
-		require.EqualError(t, err, "image:testID is not accessible under:"+server.URL+" status:404 Not Found")
+		require.EqualError(t, err, "image:\"testID\" is not accessible under:"+server.URL+", status:404 Not Found")
 	})
 
 	t.Run("Unsupported scheme", func(t *testing.T) {
 		err := checkImageURL("testID", "ftp://unsupported.url", nil)
-		require.EqualError(t, err, "image:testID with url:ftp://unsupported.url has unkown protocol")
+		require.EqualError(t, err, "image: \"testID\" with url: ftp://unsupported.url has unknown protocol")
 	})
 
 	t.Run("valid OCI URL", func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


cmd/metal-api/internal/service/image-service.go
```diff
@@ -346,15 +350,35 @@ func (r *imageResource) createImage(request *restful.Request, response *restful.
...
-func checkImageURL(id, url string) error {
-	// nolint
-	res, err := http.Head(url)
-	if err != nil {
-		return fmt.Errorf("image:%s is not accessible under:%s error:%w", id, url, err)
-	}
-	if res.StatusCode >= 400 {
-		return fmt.Errorf("image:%s is not accessible under:%s status:%s", id, url, res.Status)
+func checkImageURL(id, url, username, password string) error {
+	if strings.HasPrefix(url, "http://") || strings.HasPrefix(url, "https://") {
```
majst01: probably better to have a switch based on protocol like so:

switch protocol {
case "https://" , "http://":
  
case "oci://":

default:
 error out
}

cmd/metal-api/internal/service/partition-service.go
```diff
@@ -175,7 +175,7 @@ func (r *partitionResource) createPartition(request *restful.Request, response *
 		imageURL = *requestPayload.PartitionBootConfiguration.ImageURL
 	}
 
-	err = checkImageURL("image", imageURL)
+	err = checkImageURL("image", imageURL, "", "")
```
majst01: seems weird, having this func always with empty auth credentials ?

go.mod
```diff
@@ -38,6 +38,21 @@ require (
...
+require (
+	github.com/containerd/stargz-snapshotter/estargz v0.18.1 // indirect
+	github.com/docker/cli v29.0.3+incompatible // indirect
+	github.com/docker/distribution v2.8.3+incompatible // indirect
+	github.com/docker/docker-credential-helpers v0.9.3 // indirect
+	github.com/google/go-containerregistry v0.20.7
+	github.com/mitchellh/go-homedir v1.1.0 // indirect
+	github.com/vbatts/tar-split v0.12.2 // indirect
+)
+
```
majst01: no additional requires, and sort them into direct and indirect deps.

majst01:
test would be nice

cmd/metal-api/internal/service/image-service.go
```diff
@@ -346,15 +351,41 @@ func (r *imageResource) createImage(request *restful.Request, response *restful.
...
-func checkImageURL(id, url string) error {
-	// nolint
-	res, err := http.Head(url)
-	if err != nil {
-		return fmt.Errorf("image:%s is not accessible under:%s error:%w", id, url, err)
-	}
-	if res.StatusCode >= 400 {
-		return fmt.Errorf("image:%s is not accessible under:%s status:%s", id, url, res.Status)
+func checkImageURL(id, inputURL string, ociCredentials authn.Authenticator) error {
+	parsedURL := strings.Split(inputURL, "://")
```
vknabel: I think we should use `url.Parse` here.

cmd/metal-api/internal/service/image-service.go
```diff
@@ -346,15 +351,41 @@ func (r *imageResource) createImage(request *restful.Request, response *restful.
...
+		if ociCredentials == nil {
+			ociCredentials = authn.Anonymous
+		}
+
+		_, err = remote.Image(ref, remote.WithAuth(ociCredentials))
+		if err != nil {
+			return fmt.Errorf("image:%s is not accessible under:%s error:%w", id, inputURL, err)
+		}
+	default:
+		return fmt.Errorf("image:%s with url:%s has unkown protocol", id, inputURL)
```
vknabel:
```suggestion
		return fmt.Errorf("image: %q with url: %s has unknown protocol", id, inputURL)
```

cmd/metal-api/internal/service/image-service.go
vknabel, line 445: If I understand this correctly, `updateImage` is impossible with `oci` images that require any kind of auth. Is this intended?

cmd/metal-api/internal/service/image-service.go
```diff
@@ -346,15 +351,41 @@ func (r *imageResource) createImage(request *restful.Request, response *restful.
...
+			return fmt.Errorf("image reference:%s could not be parsed. error:%w", inputURL, err)
+		}
+
+		if ociCredentials == nil {
+			ociCredentials = authn.Anonymous
+		}
+
+		_, err = remote.Image(ref, remote.WithAuth(ociCredentials))
+		if err != nil {
+			return fmt.Errorf("image:%s is not accessible under:%s error:%w", id, inputURL, err)
```
vknabel:
```suggestion
			return fmt.Errorf("image: %q is not accessible under: %s, error: %w", id, inputURL, err)
```

cmd/metal-api/internal/service/image-service.go
```diff
@@ -346,15 +351,41 @@ func (r *imageResource) createImage(request *restful.Request, response *restful.
...
+		}
+		if res.StatusCode >= 400 {
+			return fmt.Errorf("image:%s is not accessible under:%s status:%s", id, inputURL, res.Status)
+		}
+	case "oci":
+		ref, err := name.ParseReference(parsedURL[1])
+		if err != nil {
+			return fmt.Errorf("image reference:%s could not be parsed. error:%w", inputURL, err)
```
vknabel:
```suggestion
			return fmt.Errorf("image reference: %q could not be parsed, error: %w", inputURL, err)
```

cmd/metal-api/internal/service/image-service.go
```diff
@@ -346,15 +351,41 @@ func (r *imageResource) createImage(request *restful.Request, response *restful.
...
-	if res.StatusCode >= 400 {
-		return fmt.Errorf("image:%s is not accessible under:%s status:%s", id, url, res.Status)
+func checkImageURL(id, inputURL string, ociCredentials authn.Authenticator) error {
+	parsedURL := strings.Split(inputURL, "://")
+
+	switch parsedURL[0] {
+	case "http", "https":
+		_, err := url.ParseRequestURI(inputURL)
+		if err != nil {
+			return fmt.Errorf("image:%s could not be parsed. error:%w", inputURL, err)
```
vknabel:
```suggestion
			return fmt.Errorf("image: %q could not be parsed, error: %w", inputURL, err)
```

cmd/metal-api/internal/service/image-service.go
```diff
@@ -346,15 +351,41 @@ func (r *imageResource) createImage(request *restful.Request, response *restful.
...
+		}
+
+		res, err := http.Head(inputURL)
+		if err != nil {
+			return fmt.Errorf("image:%s is not accessible under:%s error:%w", id, inputURL, err)
```
vknabel:
```suggestion
			return fmt.Errorf("image: %q is not accessible under: %s, error: %w", id, inputURL, err)
```

cmd/metal-api/internal/service/image-service.go
```diff
@@ -346,15 +351,41 @@ func (r *imageResource) createImage(request *restful.Request, response *restful.
...
+		}
+		if res.StatusCode >= 400 {
+			return fmt.Errorf("image:%s is not accessible under:%s status:%s", id, inputURL, res.Status)
```
vknabel:
```suggestion
			return fmt.Errorf("image: %q is not accessible under: %s, status: %s", id, inputURL, res.Status)
```

cmd/metal-api/internal/service/partition-service.go
```diff
@@ -175,7 +175,7 @@ func (r *partitionResource) createPartition(request *restful.Request, response *
 		imageURL = *requestPayload.PartitionBootConfiguration.ImageURL
 	}
 
-	err = checkImageURL("image", imageURL)
+	err = checkImageURL("image", imageURL, nil)
```
vknabel, line 178: Same here and within that file: is `nil` intentional?

cmd/metal-api/internal/service/image-service.go
mac641: I tried `url.Parse` but unfortunately it does not support `oci://`-based URIs.
If you know a better solution, please let me know.

cmd/metal-api/internal/service/image-service.go
mac641: To be frank, I haven't dug deeper into `updateImage`. To my understanding this is part of the `imageResource` which is part of the old installation method where tarballs had to be downloaded and extracted. We don't need it for `oci` images since we can use `github.com/google/go-containerregistry/pkg/v1/remote` to download/verify/... the images from a registry.

cmd/metal-api/internal/service/partition-service.go
mac641: Since we don't have any registry yet that needs authentication, but still want to support it, I can't think of a better way than passing over `nil` for now.

Feel free to suggest a solution that's more to your liking and I'm happy to add it.

cmd/metal-api/internal/service/image-service.go
mac641: I reworked this part. Feel free to have a look...

cmd/metal-api/internal/service/partition-service.go
vknabel: Not right now. Could you create an issue for this? Then we can tackle this in the future.

PR #643, opened 2026-03-13
Add new network properties to machineNetwork

majst01:
## Description

In Order to be able to gradually migrate to new apiv2 consumers, also put new networktypes into the machineNetwork into the Allocation.

### DIFF
--- a/cmd/metal-api/internal/metal/machine.go
+++ b/cmd/metal-api/internal/metal/machine.go
@@ -378,17 +378,20 @@ func (ms Machines) WithRole(role Role) Machines {
 
 // MachineNetwork stores the Network details of the machine
 type MachineNetwork struct {
-	NetworkID           string   `rethinkdb:"networkid" json:"networkid"`
-	Prefixes            []string `rethinkdb:"prefixes" json:"prefixes"`
-	IPs                 []string `rethinkdb:"ips" json:"ips"`
-	DestinationPrefixes []string `rethinkdb:"destinationprefixes" json:"destinationprefixes"`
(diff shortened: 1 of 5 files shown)


cmd/metal-api/internal/metal/machine.go
```diff
@@ -378,17 +378,20 @@ func (ms Machines) WithRole(role Role) Machines {
...
+	Vrf                 uint          `rethinkdb:"vrf" json:"vrf"`
+	PrivatePrimary      bool          `rethinkdb:"privateprimary" json:"privateprimary"`
+	Private             bool          `rethinkdb:"private" json:"private"`
+	ASN                 uint32        `rethinkdb:"asn" json:"asn"`
+	Nat                 bool          `rethinkdb:"nat" json:"nat"`
+	Underlay            bool          `rethinkdb:"underlay" json:"underlay"`
+	Shared              bool          `rethinkdb:"shared" json:"shared"`
+	ProjectID           string        `rethinkdb:"projectid" json:"projectid"`
+	NetworkTypeV2       NetworkTypeV2 `rethinkdb:"networktype" json:"networktype"`
+	NATTypeV2           NATType       `rethinkdb:"nattype" json:"nattype"`
```
Gerrit91: Either we need to add these new fields to metal-apiserver, too, or we can dynamically provide them and only have them in the V1 API.

Gerrit91 (approved):

cmd/metal-api/internal/metal/machine.go
Gerrit91: Decided to just carry the fields over to metal-apiserver.

PR #646, opened 2026-04-20
Migrate reserved to tainted

majst01:
## Description

This is a preparation to MEP-4 where `reserved` machines are now called `tainted` to prevent irritations with size reservations

#### Used AI-Tools ✨

- none used for generation


cmd/metal-api/internal/datastore/machine_integration_test.go
```diff
@@ -614,16 +614,16 @@ func TestRethinkStore_SearchMachines(t *testing.T) {
 		{
 			name: "search by state value reserved",
```
Gerrit91:
```suggestion
			name: "search by state value tainted",
```

majst01: done

Gerrit91 (approved):

PR #650, opened 2026-06-01
fix: only enforce machine quota when explicitly configured (avoids default to zero)

mwennrich:
## Description

Check value of quota.machine for nil.


cmd/metal-api/internal/service/machine-service.go
```diff
@@ -1221,14 +1221,16 @@ func allocateMachine(ctx context.Context, logger *slog.Logger, ds *datastore.Ret
...
 		mq := p.GetProject().GetQuotas().GetMachine()
-		maxMachines := mq.GetQuota().GetValue()
-		var actualMachines metal.Machines
-		err := ds.SearchMachines(&datastore.MachineSearchQuery{AllocationProject: &projectID, AllocationRole: &metal.RoleFirewall}, &actualMachines)
-		if err != nil {
-			return nil, err
-		}
-		if len(actualMachines) >= int(maxMachines) {
-			return nil, fmt.Errorf("project quota for machines reached max:%d", maxMachines)
+		if mq.GetQuota() != nil {
```
majst01, line 1224: I guess this is never nil, with the suggestion probably yes
```suggestion
		if mq.Quota != nil {
```

mwennrich: > I guess this is never nil, with the suggestion probably yes

```
func (x *Quota) GetQuota() *wrapperspb.Int32Value {
	if x != nil {
		return x.Quota
	}
	return nil
}
```
It's basically the same. GetQuota returns nil, if nil.

majst01 (approved):

PR #651, opened 2026-06-12
add RoomID to switches and machines

mwennrich:
## Description

Currently, when switches in a partition are located in different rooms, there is no way to distinguish between machines in separate rooms—for example, for Kubernetes anti-affinity rules. This PR introduces a machine-room tag/label for each machine, enabling the use of such labels in anti-affinity configurations."

TODO:
  - [x] tested in test
  - [x] adding roomID to existing machines

Depends on: https://github.com/metal-stack/metal-lib/pull/209

#### Used AI-Tools ✨
 FITS/Kiero-Thinking (GLM 4.7)


majst01 (approved):
@Gerrit91 should also have a look

majst01:
can you please retrofit these changes into 

- https://github.com/metal-stack/api
- https://github.com/metal-stack/metal-apiserver

in this order

PR #655, opened 2026-06-26
Add admin status to nic port

iljarotar:
## Description

The double-role of the `Actual` field of a switch port causes a bug in the metal-core that sometimes results in ports being set to DOWN by metal-core (https://github.com/metal-stack/metal-core/issues/196).

A cleaner solution would include adding a table for the switch status and remove all status fields from the switch table. Then the reports coming from metal-core will only be reflected in the switch status table while the switch table would only hold the desired state of a switch. Once metal-core is migrated to v2API, we should work on this cleaner solution. For now, this quick solution is okay, I think.

TODO

- [x]  Allow setting port status on ports that don't have a machine connection. Otherwise, this PR would make it impossible to add new machines on ports that were previously down.

### DIFF
--- a/cmd/metal-api/internal/metal/network.go
+++ b/cmd/metal-api/internal/metal/network.go
@@ -144,18 +144,6 @@ func (ns *NicState) WantState(s SwitchPortStatus) (NicState, bool) {
 			Desired: &s,
 		}, true
 	}
-	if ns.Actual == s {
-		// we want a state we already have
-		if ns.Desired != nil {
-			return NicState{
-				Actual:  s,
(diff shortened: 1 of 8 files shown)


ulrichSchreiner:
LGTM, as it is much easier to store `Actual`/`Desired` state as-is and don't do a transition from one to the other state and vice versa. 

let's hope it works :-)

mwennrich (approved):

ulrichSchreiner (approved):

majst01:
This must be retrofitted into metal-apiserver an apiv2

